PR #5, opened 2017-07-13
created the swagger-codgen generated python-sdk's

auppunda:
I will look into bravado today, but right now these are the python-sdk's generated from swagger code-gen. Instructions to build are in the README.md file. This will create the package pykov which is follows the kov api defined in the swagger.yml file


auppunda:
There is only one real commit, the other commits where created to try to figure out how to sign off properly.

auppunda:
update:to use command import swagger_client
I am figuring out how to change the name

pacogomez:
we are going to use `bravado`

PR #15, opened 2017-07-25
allow the fetching of yaml or json swagger file

auppunda:
Signed-off-by: auppunda <auppunda@github.com>


pacogomez:
- the swagger file should be distributed with the container-service-extension python module and read from the local file upon request. It shouldn't get it from GitHub repo
- we should have only one file format (json or yaml) and perform the conversion to the other format upon request
- code needs to pass the flake8 validations

Issue #19, opened 2017-08-19
the return format is json, it should return the yaml spec


pacogomez:
https://github.com/vmware/container-service-extension/blob/7e5e360a3bee8c461a68d2031c9b15daf30f9189/container_service_extension/processor.py#L126

PR #21, opened 2017-08-24
fixed yaml get swagger call

auppunda:
Signed-off-by: auppunda <auppunda@gmail.com>


auppunda:
fixes issue #19

auppunda:
to get the yaml file from the request , use response.json() method for proper formating

PR #31, opened 2017-09-01
helm script fetching with a working helm script

auppunda:


auppunda:
don't need to merge just for so that you can see the code and test it and tell me if you had any problems. The installer is their own script so that should work but let me know if they were any problems

PR #100, opened 2018-08-14
Updated component list in Installation

TheNewStellW:
Updated the component list in "Installation" section to include a list of the components with associated ports and a description of the access.

Signed-off-by: Stellios Williams <24645566+thenewstellw@users.noreply.github.com>

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -25,9 +25,13 @@ The current document covers the following `CSE` topics:
 
 The `CSE` service is designed to be installed by the vCloud Director System Administrator on a virtual machine (the `CSE` appliance) with network connectivity to the vCloud Director infrastructure where the following components access is required:
 
-- vCloud Director instance (Public Load Balancer VIP for multiple cells)
-- vCenter Server
-- AMQP Server
+**Source** | **Destination** | **Port** | **Description**
+-------|-------------|------|------------
+CSE | vCloud Director | 443 | vCloud API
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -25,9 +25,13 @@ The current document covers the following `CSE` topics:
 
 The `CSE` service is designed to be installed by the vCloud Director System Administrator on a virtual machine (the `CSE` appliance) with network connectivity to the vCloud Director infrastructure where the following components access is required:
 
-- vCloud Director instance (Public Load Balancer VIP for multiple cells)
-- vCenter Server
-- AMQP Server
+**Source** | **Destination** | **Port** | **Description**
+-------|-------------|------|------------
+CSE | vCloud Director | 443 | vCloud API
+CSE | vCenter Server | 443 | Virtual Machine Configuration
```
sahithi: Is it necessary to have ESXi hosts line? I can't think of any place where users have to specify ESXi hosts or their ports?

Also, can you eloborate on when AMQP server runs on 5671? Is this a manual step for the user?

TheNewStellW: Hey Sahithi,
I've found that CSE requires access to the host to perform guest interaction actions. The fact that CSE never asks for the port means no one would know this access is required. My deployment has all components segregated with explicit rules for access. I imagine many other deployers of CSE will have the same design.

AMQPS (TLS/SSL encrypted AMQP) runs by default on 5671.

PR #108, opened 2018-09-13
[VCDA-755] Ensure templates catalog is always shared

andrew-ni:
Before this fix, if specified catalog already existed, it wouldn't
be shared with other orgs. If specified catalog did not exist,
it would be created, but would not be shared. CSE installation
would crash because we did not reload the org representation after
the POST call.

This fix creates the catalog if needed, always shares the catalog
to other orgs, and properly reloads the org representation

Tested with all CSE commands
@sahithi

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/108)

### DIFF
--- a/container_service_extension/config.py
+++ b/container_service_extension/config.py
@@ -19,6 +19,7 @@ from pyvcloud.vcd.client import BasicLoginCredentials
 from pyvcloud.vcd.client import Client
 from pyvcloud.vcd.client import QueryResultFormat
 from pyvcloud.vcd.client import SIZE_1MB
+from pyvcloud.vcd.exceptions import EntityNotFoundException
 from pyvcloud.vcd.org import Org
 from pyvcloud.vcd.platform import Platform
 from pyvcloud.vcd.vapp import VApp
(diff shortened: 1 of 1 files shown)


andrew-ni:
> "we tried to GET the catalog before the POST catalog creation finished." - As far as I remember, catalog creation operation is synchronus. The REST response has the catalog id. So, if both opeartion are performed in the same thread (which is the case for pyvcloud, vcd-cli and cse) there is a _happens after_ relation established between the two REST calls.

Hmm that's weird. I reached this conclusion when testing out this scenario: We don't have a `cse` catalog, so we need to create a catalog, then share it (using `Org.create_catalog()` and then `Org.share_catalog()`. `Org.share_catalog()` makes a GET call using `Org.get_catalog()`, which is where we received an `EntityNotFoundException`, resulting in the installation crashing. However, the catalog would still be created, and at this point, if we call `Org.share_catalog()`, everything proceeds as expected. Also, `Org.reload()` documentation states that it should be called after every state-changing operation made to an organization. Adding this call in between creation/sharing seems to have solved the issue.

rocknes:
Well, unless a catalog is shared, only the creator of the catalog, org admin of the org where the catalog is created and sys admin can see it. sharing the catlog makes it visible to other users. Maybe that's the core problem.

andrew-ni:
Edited!

andrew-ni:
Sorry. The confusion was that I thought the problem was that we tried to GET a catalog before the POST catalog creation finished. However, the problem is that whenever we use a method that modifies the org representation (such as POST catalog creation), we have to use `Org.reload()`.

PR #109, opened 2018-09-14
[Doc] Privileges required for VCD service account & CSE-VCD compatibility matrix

sahithi:
Documentation on below two items:

1. Minimum privileges required to create a VCD service account
2. CSE-vCD compatibility matrix

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/109)


andrew-ni:
I think for consistency it'd be nice to have the *broker group properties* be laid out in the same way that you did for *vcd group properties*.

For the broker group section: could we do *Group 'broker' has the following key properties*, and then have the following table be a 2-column table with headers *Property* and *Value*, similar to how the vcd group properties table is laid out.

Everything else looks good!!

PR #110, opened 2018-09-19
Skip AMQP configuration if current settings are same as config file settings

andrew-ni:
Minor refactoring to configure_amqp_settings() to make settings display more consistent, when showing the user current settings and the settings listed in their config file. Also added a redundancy check, so that user is only prompted to configure amqp if current settings are different than desired settings.

@sahithi

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/110)


andrew-ni:
Sorry, I've tested cse installation, and verified that amqp connection exists. I've tried using different amqp settings and was able to successfully configure vCD amqp settings during cse installation.

PR #114, opened 2018-09-24
[VCDA-745] Add --ssh-key option for cse install

andrew-ni:
We disabled root ssh password access in ubuntu VMs to follow expected behavior of ubuntu
VM templates, so root is only accessible via ssh key.

`--no-capture` option or `cleanup: false` in **config.yaml** allowed users to
access the temporary vApp and debug. Without password access, these were
inaccessible. To enable ssh access again, we've added `--ssh-key` option 
to `cse install` command. The ssh-key provided will be saved in the 
template vApp and all VMs created from the template.

Updated documentation for CSE install options.

Tested with all CSE commands
- `--no-capture` option requires `--ssh-key`
- cluster/temp VMs are accessible via ssh key if ssh-key was provided during installation
- (ubuntu) root@host is no longer accessible via password, only ssh key
- (ubuntu and photon) user@host is accessible via password and ssh key
- (photon) root@host is accessible via password and ssh key

@sahithi

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/114)


andrew-ni:
Yes, the ssh-key write into /root/.ssh/authorized_users works on photon as well. Tested this with the temporary vApp as well as cluster VMs. On photon we allow password access, so both password and public key access works.

`--ssh-key` option works for enabling ssh key root access in ubuntu and photon vms.

I'm checking with Yassine on whether we want to disable root password access for photon as well, as it was only specified that ubuntu templates had this expected behavior.

PR #119, opened 2018-09-28
Refactor config validation process

andrew-ni:
Config file validation now happens naturally when using
get_config(). Before, get_config() only loaded config.yaml into
a dict, while check_config() did all the validation. However, both
returned the config file. Refactored processing/validation steps
into focused functions for improved readability and reuse.

check_config() is renamed check_cse() and has a more defined role
of verifying that CSE has been installed correctly (checking
catalog and templates).

Updated various functions that get the config dict using
get_config(), and check_cse() is only called when it's necessary
to verify that CSE is installed correctly on vCD.

Moved validate_broker_config_content() out of broker.py into
config.py and renamed to validate_templates(). This way, everything
related to config file processing is in config.py.

Updated documentation/docstrings to match changes, minor code
cleanup, and tested with all CSE commands.

@sahithi


andrew-ni:
Just for tracking purposes: going to close this pull request. Changes will be going into a future pull request that refactors config file validation

PR #127, opened 2018-10-08
[VCD-743] Documentation fix on incorrect command usage example

sakthisunda:
- Documentation fix for viewing node info on a node with given cluster name

-  The usage example for cse node info is corrected. Verified in the installation that the command usage appears correctly and the usage works fine when gets executed in the local setup.

-  @sahithi @andrew-ni

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/127)


sakthisunda:
Suggested change by @rocknes commited.

PR #128, opened 2018-10-09
Updated help text during cluster creation

mgaruccio:
The final line of help text after creating a cluster is missing the --config switch from it's example starting command, this change just adds the switch to avoid confusion.


andrew-ni:
Hi, this is going to be fixed in an upcoming internal pull request to refactor `install_cse()`

https://github.com/andrew-ni/container-service-extension/blob/install-cse/container_service_extension/config.py

PR #137, opened 2018-10-22
[VCDA-799] Remove login requirement from all vcd cse help commands

harshneelmore:
- This fix will display help commands with no user login required. Manually tested help commands for various users[admininstrator, vappAuthor].

- @rocknes @sompa @sahithi @sakthisunda @andrew-ni

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/137)

### DIFF
--- a/container_service_extension/client/cse.py
+++ b/container_service_extension/client/cse.py
@@ -32,7 +32,8 @@ def cse(ctx):
     """
     if ctx.invoked_subcommand is not None:
         try:
-            restore_session(ctx)
+            if ctx.invoked_subcommand != 'version':
+                restore_session(ctx)
             if ctx.invoked_subcommand not in ['system', 'template', 'version']:
                 if not ctx.obj['profiles'].get('vdc_in_use') or \
                    not ctx.obj['profiles'].get('vdc_href'):
--- a/container_service_extension/client/cse.py
+++ b/container_service_extension/client/cse.py
@@ -30,17 +30,7 @@ def cse(ctx):
             Display CSE version. If CSE version is displayed, then vcd-cli has
             been properly configured to run CSE commands.
     """
-    if ctx.invoked_subcommand is not None:
-        try:
-            if ctx.invoked_subcommand != 'version':
-                restore_session(ctx)
-            if ctx.invoked_subcommand not in ['system', 'template', 'version']:
-                if not ctx.obj['profiles'].get('vdc_in_use') or \
-                   not ctx.obj['profiles'].get('vdc_href'):
-                    raise Exception('select a virtual datacenter')
-        except Exception as e:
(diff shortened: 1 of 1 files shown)


sompa:
___
*[container_service_extension/client/cse.py, line 114 at r1](https://reviewable.io/reviews/vmware/container-service-extension/137#-LPS7u8SBMCm2pc5ShZf:-LPS7u8SBMCm2pc5ShZg:b2cwtph) ([raw file](https://github.com/vmware/container-service-extension/blob/6c2768217be93c5de9e02184b70159ca3a80e138/container_service_extension/client/cse.py#L114)):*
> ```Python
> def list_templates(ctx):
>     try:
>         restore_session(ctx)
> ```

Don't you have to put the try/except block here to check if a vdc is selected ? Applies to the other places as well wherever restore_session is called.

sompa:
:lgtm:

sompa:
@harshneelmore I am somehow not seeing the option to reopen my comment, however after an offline discussion with @rocknes the check for vdc selection is indeed needed at the cluster level. @rocknes please feel free to elaborate on this one.

harshneelmore:
___
*[container_service_extension/client/cse.py, line 86 at r1](https://reviewable.io/reviews/vmware/container-service-extension/137#-LPSzrgl3dMTtoe524E-:-LPT1Ra51GCEdNJ-APc2:b-sitqfc) ([raw file](https://github.com/vmware/container-service-extension/blob/22d17932e1e000b15bfa41a3de922c957493993a/container_service_extension/client/cse.py#L86)):*

I partially agree with Sompa. Restore_session in itself doesn't do much more than read profiles.yaml, log in with the read credentials, and create a Client object. So the try catch block isn't necessarily needed wherever restore_session is called.

However we shouldn't completely get rid of the try-catch block. I believe these try-catch blocks were put in place to fail fast and standardize the error messages for commands that rely on availibility of orgvdc.

From the looks of this if statement the try catch block applies to only the ``vcd cse cluster...`` and ``vcd cse node..`` commands and can be pushed down to those individual commands.

Btw can you login into system org as sys admin and run one of the vcd cse cluster command and paste the current output after your fix?
</blockquote></details>

Following is the output:

(cse) moreh-a01:container-service-extension moreh$ vcd login 10.150.202.255 system administrator -iwp ca\$hc0w
administrator logged in, org: 'system', vdc: ''
(cse) moreh-a01:container-service-extension moreh$ vcd cse cluster -h
Usage: vcd cse cluster [OPTIONS] COMMAND [ARGS]...

  Work with kubernetes clusters.

Options:
  -h, --help  Show this message and exit.

Commands:
  config  get cluster config
  create  create cluster
  delete  delete cluster
  info    get cluster info
  list    list clusters

PR #141, opened 2018-10-25
[VCDA-789] Refactor CSE Installation - AMQP, registration, catalog

andrew-ni:
Generalized AMQP configuration and CSE registration
- `install_cse()` uses `should_configure_amqp()` and `should_register_cse()` to decide whether or not to configure amqp settings or register extension based on existing settings and user input

Cleaned up `config.py` and `utils.py`
- `config.py` should only include code specific to the config file and cse installation
- `utils.py` should include general use code
- General functions `get_sha256()`, `bool_to_msg()`, `configure_vcd_amqp()`, `create_amqp_exchange()`, `register_extension()` have all been moved to `utils.py`
- AMQP configuration and CSE registration has been moved to occur before template creation, since template creation takes the longest and is most prone to failure. This way, CSE installation can just finish after template creation.
- Moved catalog creation/sharing up to isolate from template creation step

Tested CSE installation :)

@rocknes @sahithi

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/141)

### DIFF
--- a/container_service_extension/config.py
+++ b/container_service_extension/config.py
@@ -27,7 +27,6 @@ from vcd_cli.utils import to_dict
 from vsphere_guest_run.vsphere import VSphere
 import yaml
 
-from container_service_extension.consumer import EXCHANGE_TYPE
 from container_service_extension.utils import bool_to_msg
 from container_service_extension.utils import catalog_exists
 from container_service_extension.utils import catalog_item_exists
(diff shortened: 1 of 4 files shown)


andrew-ni:
Decided to move extension registration and AMQP configuration responsibilities from **utils.py** back to **config.py**, because we should only ever be using those functions during CSE installation.

PR #172, opened 2018-12-06
[VCDA-817] Implement rollback for node creation failure

harshneelmore:
- Implemented rollback for node creation failures.
- Refactored rollback_decorator to combine cluster creation and node creation.
- Tested the above implementation manually by creating faulty clusters and nodes.

- @sakthisunda @sahithi @andrew-ni

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/172)

### DIFF
--- a/container_service_extension/broker.py
+++ b/container_service_extension/broker.py
@@ -791,13 +791,13 @@ class DefaultBroker(threading.Thread):
         template = self.get_template()
         try:
             delete_nodes_from_cluster(self.config, vapp, template,node_list, force=True)
-        except Exception as e:
-            LOGGER.error("Couldn't delete node %s from cluster:%s" % (node_list, self.cluster_name))
+        except Exception:
+            LOGGER.warning("Couldn't delete node %s from cluster:%s" % (node_list, self.cluster_name))
         for vm_name in node_list:
             vm = VM(self.client_tenant, resource=vapp.get_vm(vm_name))
             try:
                 vm.undeploy()
-            except Exception as e:
+            except Exception:
                LOGGER.warning("Couldn't undeploy VM %s" % vm_name)
         vapp.delete_vms(node_list)
         LOGGER.debug('Successfully deleted nodes: %s' % node_list)
--- a/container_service_extension/broker.py
+++ b/container_service_extension/broker.py
@@ -39,10 +39,12 @@ from container_service_extension.exceptions import CseServerError
 from container_service_extension.exceptions import MasterNodeCreationError
 from container_service_extension.exceptions import NFSNodeCreationError
 from container_service_extension.exceptions import WorkerNodeCreationError
+from container_service_extension.exceptions import NodeCreationError
 from container_service_extension.utils import ERROR_DESCRIPTION
 from container_service_extension.utils import ERROR_MESSAGE
 from container_service_extension.utils import SYSTEM_ORG_NAME
 from container_service_extension.utils import error_to_json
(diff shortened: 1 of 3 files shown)


harshneelmore:
___
*[container_service_extension/broker.py, line 795 at r2](https://reviewable.io/reviews/vmware/container-service-extension/172#-LT9ITMxAV4DDj-kMZSJ:-LTPSJs58pEVUpHXf99_:bc865v4) ([raw file](https://github.com/vmware/container-service-extension/blob/28aaf1561c9e86a8f1c990e6f30011526ab2efa4/container_service_extension/broker.py#L795)):*

Just curious. As this is right now, we try to undeploy the VM, and if that gives us an exception, we just move on to the next step right? Do we know what cases throw an exception here?
</blockquote></details>

The exception can be caused if the VM is configured incorrectly. Even if the undeploy fails we can still delete the VM. Undeploy and then delete is a clean way to remove the VM form vApp.

harshneelmore:
*[a discussion](https://reviewable.io/reviews/vmware/container-service-extension/172#-LTPWPfb7r1kVGv5DE_n:-LTPbNpv2eb9jowQpo2O:bnnxm9m) (no related file):*
<details><summary><i>Previously, hodgesrm (Robert Hodges) wrote…</i></summary><blockquote>

Even if we defer tests we still need to understand how they will work.  Looking at the code it's not obvious what would be the best way to do verification, which is a sign that we might need to rethink the design a bit.  For example, the chaos-monkey style of test might require changing the CSE server to inject failures.  
</blockquote></details>

@hodgesrm For current testing, I have manually done the following 
- create a cluster with the wrong config, 100 cpu's 
- create 15 nodes, nodes with the wrong network.
So far these were the cases I could identify but  later we can add a testing framework around rollback failures

---

harshneelmore:
___
*[container_service_extension/broker.py, line 814 at r2](https://reviewable.io/reviews/vmware/container-service-extension/172#-LT9CSJs7SfqHocFQ7Ck:-LTPcedf08z2awEViR1M:b-tpcmdo) ([raw file](https://github.com/vmware/container-service-extension/blob/28aaf1561c9e86a8f1c990e6f30011526ab2efa4/container_service_extension/broker.py#L814)):*

This is an implementation detail from pyvcloud that we need to be aware of, and i ran into the same problems working with installation code.

Whenever we use a VDC method that changes the representation of the vdc in vCD, we should call `vdc.reload()` afterwards. However, delete_vapp is an async operation, so to properly wait for it to complete, we should be using `stdout()` from `vcd_cli.utils.py`. However, `that stdout()` function requires a click.core.Context `ctx` parameter that we don't have access to in this function.

Ideally the code for this action should look like:
```python
task = vdc.delete_vapp(self.cluster['name'], force=True)
stdout(task, ctx=ctx)
vdc.reload()
```

For now, I think it's best to use this, until we can make a function like vcd_cli.utils.stdout(), but without the ctx object parameter:
```python
task = vdc.delete_vapp(self.cluster['name'], force=True)
time.sleep(15)  # import time, sleep for 15 sec, should be enough time for vapp to be deleted
vdc.reload()
```

Check in config.py where we use vapp, vdc, org functions and how we wait and reload.

Of if you intend for this to be asynchronous and you won't use the VDC object afterwards, then you can just add an inline comment describing that the async behavior is intended.
</blockquote></details>

Will talk to @sahithi  and see if the above is needed

PR #178, opened 2018-12-13
[VCDA-834] Implement flag to disable rollback for troubleshooting

harshneelmore:
- By default, rollback is enabled to disable rollback user can add the flag **--disable-rollback** to cluster/node creation.

- Tested the implementation as follows:
Created faulty clusters/nodes and allowing rollback to trigger cleanup.
Created faulty clusters/nodes with rollback disabled to not trigger cleanup.

- @sakthisunda @andrew-ni @sahithi @hodgesrm

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/178)

### DIFF
--- a/container_service_extension/broker.py
+++ b/container_service_extension/broker.py
@@ -40,7 +40,6 @@ from container_service_extension.exceptions import MasterNodeCreationError
 from container_service_extension.exceptions import NFSNodeCreationError
 from container_service_extension.exceptions import NodeCreationError
 from container_service_extension.exceptions import WorkerNodeCreationError
-from container_service_extension.exceptions import NodeCreationError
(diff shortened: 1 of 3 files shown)


harshneelmore:
___
*[container_service_extension/client/cluster.py, line 90 at r1](https://reviewable.io/reviews/vmware/container-service-extension/178#-LTxVmkeDTk61YeouyF0:-LTylD_30Fn_c_n6jS0M:bty5piz) ([raw file](https://github.com/vmware/container-service-extension/blob/013afbbe52890c6a500de39fcfda8280c1cb2101/container_service_extension/client/cluster.py#L90)):*

nit: fullstop at the end.
</blockquote></details>

Removed the fullstop

PR #194, opened 2019-01-09
VCDA-858: PKS and Uaac Client; Sample code to establish connection with PKS

sahithi:
This PR includes below items

- Swagger generated PKS client.
- Modified Uaac client (from https://github.com/dattnguyen82/PyUaaClient/blob/master/PyUaaClient.py) to work with python 3.7
- Sample prototype code (connection.py) to establish connection with pks. This file will be eventually removed once we have pks broker in place.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/194)


sompa:
___
*[container_service_extension/utils.py, line 668 at r2](https://reviewable.io/reviews/vmware/container-service-extension/194#-LWO7EgECWr0IcT313dx:-LWOGVdZApefebdRfffw:b-bg8s3c) ([raw file](https://github.com/vmware/container-service-extension/blob/3c914c80daaea7ae565b60da9739dbb810507262/container_service_extension/utils.py#L668)):*

If I remember correctly vcd-cli has the exact same piece of code. Probably we should move this to pyvcloud at some point.
</blockquote></details>

Agreed. We should plan to move this util method into pyvcloud util instead and keep cse as light as possible.

PR #206, opened 2019-01-25
[VCDA-865] Remove vCD AMQP configuration and Enhance Extension Checks

### LINKED ISSUE #187: vCloud AMQP Exchange type
We use our own system of automation which use routing on RabbitMQ Exchange level. But if we try install cse, we have error about type of RabbitMQ Exchange - pika.exceptions.ChannelClosed: (406, "PRECONDITION_FAILED - inequivalent arg 'type' for exchange 'vcdExchange' in vhost '/': received 'direct' but current is 'topic'").  We use topic type of Exchange for routing.
If we try use alternate-exchange on RabbitMQ Exchange we also get an error - pika.exceptions.ChannelClosed: (406, "PRECONDITION_FAILED - inequivalent arg 'alternate-exchange' for exchange 'vcdExchange' in vhost '/': received none but current is the value 'vcdExchangeAlternative' of type 'longstr'")

In this case, we cannot organize message routing for vCloud Events for example.

andrew-ni:
* We were configuring vCD AMQP settings during CSE installation when we should not. If other extensions are being used, this re-configuring could cause errors in those extensions. This functionality has been removed. (Related to #187)

* Fix a bug where if CSE is already registered as an extension to vCD, but has different properties than the config file, then we do not update vCD-CSE properties. Now when installing CSE, if CSE already exists on vCD but has different properties than config file, the user is alerted and presented with an option to update CSE properties on vCD. (Dependent on vmware/pyvcloud#379)

* Remove system testing functionality of checking vCD AMQP configuration

* Update docs to reflect that we do not configure vCD AMQP anymore

* Fixes #187 

* System tests pass

@rocknes

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/206)


sompa:
:lgtm:

PR #207, opened 2019-01-26
[VCDA-843] Update CSE install method to register new Right for CSE authorization

harshneelmore:
Signed-off-by: Harshneel More <moreh@vmware.com>

- Add support in CSE install method to register a new right for CSE 
- Manually tested by running a new CSE install.
- Manually tested CSE install with upgrade flag.

- @rocknes @andrew-ni @sahithi

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/207)


harshneelmore:
___
*[container_service_extension/config.py, line 1101 at r1](https://reviewable.io/reviews/vmware/container-service-extension/207#-LXLN2gk1begrwpef19Z:-LXLNrK-4aYC9iti233t:brd7h62) ([raw file](https://github.com/vmware/container-service-extension/blob/eb717dec0ac1672a4143d9fe08e371b70f3b71c7/container_service_extension/config.py#L1101)):*

Why are we raising an exception if rights have already been registered? Just ignore and move on.
</blockquote></details>

Currently, if the right is already registered then VCD gives a 400 Bad request exception, We need to check if the 400 exception is related to the right already existing if it is then we just move on and display the message "Right already exists in vCD", If its any other exception apart from Right already exists then its raised.

PR #224, opened 2019-02-08
Add missing requirement for click

gmcelhoe:
The code requires click>=7.0 as a requirement

Discovered this while installing the scripts in an empty python container.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/224)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 cachetools >= 2.0.1
-click >= 7.0
+click >= 6.7
 humanfriendly >= 4.8
 pika >= 0.11.2
 pyvcloud >= 20.0.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 cachetools >= 2.0.1
+click >= 7.0
 humanfriendly >= 4.8
 pika >= 0.11.2
 pyvcloud >= 20.0.1


rocknes:
The click requirement should have been met by vcd-cli. (click >= 6.7)
CSE has a requirement on vcd-cli.

Can you please elaborate on what exactly in CSE depends on click > 7.0?

gmcelhoe:
I count at least 7 different python files in CSE that import click.
https://github.com/vmware/container-service-extension/search?q=import+click&unscoped_q=import+click

It is good practice not to rely on transitive dependencies; what if vcd-cli changes?

rocknes (commented):
I understand your concern. CSE distribution has both the server and client rolled into one piece of software. CSE client is a plugin for vcd-cli. And both of them rely heavily on click. So the transitive dependency is intentional. We want to use the same version of click as vcd-cli.

However, the other parts of CSE (the testing framework/ the server side cli) where we are using click independent of vcd-cli, should be decoupled from vcd-cli's click dependency.

So, let's add the dependency on click explicitly in CSE but let's not introduce stricter version requirement than vcd-cli.

> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/224#-:-LYTjkgj14hGg_k0_O8k:b-rztdld)* status: 0 of 1 files reviewed, all discussions resolved

PR #230, opened 2019-02-11
Improve check handling of bad VC passwords

gmcelhoe:
If the VC password is bad, nothing is printed to the console except that the
config file was invalid.

This change adds a message, explaining the problem.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/230)


andrew-ni:
Hey Glenn,

I actually have a task this sprint that improves check handling for the majority of missing vCD entity errors (missing org, ovdc, external network, etc). All of these errors also result in the misleading *config file is invalid* message.

gmcelhoe:
> I actually have a task this sprint that improves check handling for the majority of missing vCD entity errors (missing org, ovdc, external network, etc). All of these errors also result in the misleading _config file is invalid_ message.

PR #248, opened 2019-02-20
VCDA-902 : Implementation of  CPI id, developer-mode, removal of rp-paths

sompa:
- Implementation of  CPI id, developer-mode, removal of rp-paths
- This PR implements logic to include CPI (Cloud Provider Id) associated with each datacenter in VC, in the new PKS configs. Removes the resource pool path field from pvdcs and adds datacenter and cluster field instead. Also adds logic to include a developer mode field 'pks_proxy_ip' to work with PKS dev set up where the PKS and VC reside in a different network than vCD's.

- @sahithi @sakthisunda @rocknes @andrew-ni

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/248)


sompa:
___
*[container_service_extension/configure_cse.py, line 68 at r8](https://reviewable.io/reviews/vmware/container-service-extension/248#-LZBVtUm5BNT6ILGwJGM:-LZCGNSR6-wSWDUqYWgr:bm9syxa) ([raw file](https://github.com/vmware/container-service-extension/blob/bb267c7aa94859dc099f6d6dd14a1a3e25318e1e/container_service_extension/configure_cse.py#L68)):*

Mention in the instructions of pks_accounts that users can add a key-value pair (proxy) and what it means.
</blockquote></details>

imho this should not be part of the instructions, as only developers have access to this information, may be we can document it as part of our contributing guide but not in the actual instructions of filling out this config file.

sompa:
> The commit history for this PR is a little muddy. Do you know why is it so?
> Btw please use -s flag while committing.  That helps us to avoid the 'dco required' tag.

Yes, I had a rebase done that did not go through completely and left my fork master in a state where every new merge commit is appearing as a new commit on both my master and any branch that gets branched off my forked master. I was able to hard reset my master, however, I think the branch can just be discarded and a new fresh PR can be created based of a new branch (since there are 30+ commits that I highly think will get messed up if try to replay them on my current branch)

PR #297, opened 2019-04-10
Update PKS client package with new version of PKS

sahithi:
- Generated swagger code for two separate specs of pks (v1 and v1beta) and updated pksclient package
- Created separate packages v1 and v1beta under each package of api and models.
- v1 endpoint supports all operations except CRUD of compute profiles and creation of cluster with compute profile.
- v1beta endpoint supports CRUD of compute profiles, Creation and Get-cluster info on cluster only, nothing more.

Testing:
Basic sanity testing of PKS workflows only.

Note: list-clusters may not function as expected with the chang as PKS team missed out adding support for it in v1beta endpoint. So, list on clusters created with compute-profiles will not work as before. Fix will be in soon.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/297)


sompa:
Just mention a note for the commented code in the __init__ files in the new v1 and v1beta packages that its intentional.

PR #314, opened 2019-04-25
Add interview command

gmcelhoe:
Add a command which will walk through the configuration asking
questions and offering suggestions at each stage.

This is still work-in-progress, it will walk through the interview,
but still has some UX issues, and only prints the result YAML
to the console.

Unit tests are included, but system tests are not written yet.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/314)

### DIFF
--- a/container_service_extension/cse.py
+++ b/container_service_extension/cse.py
@@ -3,6 +3,7 @@
 # container-service-extension
 # Copyright (c) 2017 VMware, Inc. All Rights Reserved.
 # SPDX-License-Identifier: BSD-2-Clause
+import traceback
 
 import click
 from pyvcloud.vcd.exceptions import EntityNotFoundException
@@ -17,6 +18,7 @@ from container_service_extension.configure_cse import generate_sample_config
 from container_service_extension.configure_cse import get_validated_config
 from container_service_extension.configure_cse import install_cse
 from container_service_extension.exceptions import AmqpConnectionError
+from container_service_extension.questions import cse_interview
 from container_service_extension.service import Service
 
 CONTEXT_SETTINGS = dict(help_option_names=['-h', '--help'])
@@ -189,6 +191,34 @@ def check(ctx, config, check_install, template):
         check_cse_installation(config_dict, check_template=template)
     except EntityNotFoundException:
         click.secho("CSE installation is invalid", fg='red')
+
+
+@cli.command(short_help="Runs an interview to configure the extension.")
+@click.pass_context
+@click.option(
+    '-c',
+    '--config',
+    'config',
+    type=click.Path(exists=True),
+    metavar='<config-file>',
+    envvar='CSE_CONFIG',
+    default='/var/config.yaml',
+    help='Config file to use.')
+@click.option(
+    '-d',
+    '--detailed',
+    'detailed',
+    is_flag=True,
+    required=False,
+    default=False,
+    help='Checks that CSE is installed on vCD according to the config file')
+def interview(ctx, config, detailed):
+    """Interview for CSE configuration."""
+    try:
+        cse_interview(config, detailed)
+    except Exception as ex:
+        click.secho(f"Error performing interview. {str(ex)}", fg='red')
+        traceback.print_exc()
 
 
 @cli.command(short_help='install CSE on vCD')
--- a/container_service_extension/interview/__init__.py
+++ b/container_service_extension/interview/__init__.py
@@ -1,0 +1,20 @@
+# container-service-extension
+# Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+# SPDX-License-Identifier: BSD-2-Clause
+import traceback
+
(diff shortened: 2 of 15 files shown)


rocknes (changes requested):
> Reviewed 2 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/314)* status: 2 of 15 files reviewed, 5 unresolved discussions (waiting on @gmcelhoe)

___
*[container_service_extension/cse.py, line 221 at r1](https://reviewable.io/reviews/vmware/container-service-extension/314#-LdkQ3KcBJgxQSEx6U8d:-LdkQ3Kd6eK1iDVmx2p6:b6url49) ([raw file](https://github.com/vmware/container-service-extension/blob/7b6bbc851ae49457a7f51bee15dbe000f1434fb1/container_service_extension/cse.py#L221)):*
> ```Python
>     except Exception as ex:
>         click.secho(f"Error performing interview. {str(ex)}", fg='red')
>         traceback.print_exc()
> ```

We should pipe the exception into the client debug log, and avoid it from showing up on the console
___
*[container_service_extension/questions.py, line 49 at r1](https://reviewable.io/reviews/vmware/container-service-extension/314#-LdkqyV93VDROdZMRrEo:-LdkqyV93VDROdZMRrEp:b-jlyytn) ([raw file](https://github.com/vmware/container-service-extension/blob/7b6bbc851ae49457a7f51bee15dbe000f1434fb1/container_service_extension/questions.py#L49)):*
> ```Python
> from container_service_extension.utils import get_org
> 
> SAMPLE_TEMPLATE_PHOTON_V2 = {
> ```

Why aren't we simply importing this from configure_cse.py ?
___
*[container_service_extension/questions.py, line 118 at r1](https://reviewable.io/reviews/vmware/container-service-extension/314#-LdkrncYCbXY6GpR0oHz:-LdkrncYCbXY6GpR0oI-:b-9y1ugc) ([raw file](https://github.com/vmware/container-service-extension/blob/7b6bbc851ae49457a7f51bee15dbe000f1434fb1/container_service_extension/questions.py#L118)):*
> ```Python
>     org.reload()
> 
>     if power_on:
> ```

nit: should_power_on?
___
*[container_service_extension/questions.py, line 187 at r1](https://reviewable.io/reviews/vmware/container-service-extension/314#-LdoUlg_1QaWEdN3uXHk:-LdoUlg_1QaWEdN3uXHl:bplinkb) ([raw file](https://github.com/vmware/container-service-extension/blob/7b6bbc851ae49457a7f51bee15dbe000f1434fb1/container_service_extension/questions.py#L187)):*
> ```Python
> def load_vcd_versions(question):
>     """
>     Log into the vCD host, and configure the context with the authorized client
> ```

This docstring will fail flake8 validation.
Missing period at end of first line, and first line being empty.
___
*[container_service_extension/questions.py, line 194 at r1](https://reviewable.io/reviews/vmware/container-service-extension/314#-LdoV08IE0QY6baHdXbF:-LdoV08J2SixXiOKXeeC:b9e2pkd) ([raw file](https://github.com/vmware/container-service-extension/blob/7b6bbc851ae49457a7f51bee15dbe000f1434fb1/container_service_extension/questions.py#L194)):*
> ```Python
>                         verify_ssl_certs=str2bool(
>                             question.context.get_yaml_value('verify')))
>     versions = vcd_client.get_supported_versions_list()
> ```

We should logout the client after it's utility is over.

andrew-ni:
I've looked over this a few times now and I think the design has been well thought out.

Can you give us an example code flow of the interview noninteractive as well as interactive (walk us through something like interviewing for the vcd config file group).
Can you show us how the interview handles all objects it encounters in the stack?
    - Group, Comment, Question, Confirmation
    - When/why InteractionContexts and TopLevelContext come into play and what they do/what they're used for
    - Derived class vs parent class attributes/methods that are used during the interview
I think this would be valuable to help our understanding of these object relations and inheritances (especially the multiple inheritance objects)

InteractionContext is confusing to me. I'm understanding it as it's a class that is partly static/global while also partly local to the concrete Group/IndexedGroup that derives from it?
basically keeps track of positioning while inside of a group/entire config file?
It seems to directs question flow as well as be the hub to retrieve common values independent to any single question's inherent context?
Could you explain what InteractionContext's role is/some of its responsibilities/when it's used during the interview process?

I'm also confused what self.skip_function is intended to do/skip, as it's only ever used in Comment class. Question never uses it, but it's explicitly set in the Question constructors.



Few other small things:

I think a few of these names could be made more intuitive:
Confirmation -> InterviewConfirmation, since this is a special case question at the end of the interview
TopLevelContext -> InterviewContext ??? unsure about this one, but TopLevelContext was confusing for me

We shouldn't use `assert` in program code, because it's intended for testing purposes, and any asserts are stripped from the code when python is run with the -O flag

PR #332, opened 2019-05-06
Avail plans to only sys admins and removing the org vdc in use to be passed as a param for vdc to resize clusters.

sompa:
Avail plans to only sys admins and removing the org vdc in use to be passed as a param to resize clusters.

Signed-off-by: Sompa Malakar <malakars@vmware.com>

- This PR fixes the pks plans command option added to `vcd cse ovdc list` to be only available to system administrators. Also removes the org vdc in use from profiles to be passed as a param for vdc, if user does not specify a vdc as an option explicitly. 

- Tested `vcd cse ovdc list -p` as sys admin, org admin and vappauthor. Plan information is available only to system administrators.
- @sahithi @andrew-ni

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/332)

### DIFF
--- a/container_service_extension/broker_manager.py
+++ b/container_service_extension/broker_manager.py
@@ -9,7 +9,7 @@ from http import HTTPStatus
 
 from pyvcloud.vcd.org import Org
 
-from container_service_extension.exceptions import ClusterNotFoundError
+from container_service_extension.exceptions import ClusterNotFoundError, UnauthorizaedActionError
 from container_service_extension.exceptions import CseServerError
 from container_service_extension.exceptions import PksServerError
 from container_service_extension.logger import SERVER_LOGGER as LOGGER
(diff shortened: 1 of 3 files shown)


sompa:
___
*[container_service_extension/broker_manager.py, line 193 at r1](https://reviewable.io/reviews/vmware/container-service-extension/332#-LeJ6Hz5E7A7pOa_YNx0:-LeJNZ9G0GI0BHLfa5eX:b7ofcje) ([raw file](https://github.com/vmware/container-service-extension/blob/6defe5f0a8b9c69f6e11a04d7547b222aae5c973/container_service_extension/broker_manager.py#L193)):*

Do we have a better sub class of exception to raise?
</blockquote></details>

We currently don't have any. Infact, authorization class itself uses this plain Exception in the implementation of the secure decorator. Let me know if this still needs to change to anything else.

PR #350, opened 2019-05-30
[VCDA-1037] vcd cse node info command broken after enabling PKS

harshneelmore:
Added the param to include ORGVDC in use to the request parameters for node info.
- Tested the above change with native for Org Admin user.
- Tested the changes after enabling PKS for Org Admin user.

@sakthisunda @rocknes @sahithi

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/350)

### DIFF
--- a/container_service_extension/broker_manager.py
+++ b/container_service_extension/broker_manager.py
@@ -55,6 +55,7 @@ class Operation(Enum):
     ENABLE_OVDC = 'enable ovdc'
     INFO_OVDC = 'info ovdc'
     GET_CLUSTER_CONFIG = 'get cluster config'
+    GET_NODE_INFO = 'get node info'
 
 
 class BrokerManager(object):
@@ -173,6 +174,11 @@ class BrokerManager(object):
             list_pks_plans = self.req_spec.get('list_pks_plans', False)
             result['body'] = self._list_ovdcs(
                 list_pks_plans=list_pks_plans)
+        elif op == Operation.GET_NODE_INFO:
+            node_spec = \
+                {'cluster_name': self.req_spec.get('cluster_name', None),
+                 'node_name': self.req_spec.get('node_name', None)}
+            result['body'] = self._get_node_info(**node_spec)[0]
 
         return result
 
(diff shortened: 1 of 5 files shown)


harshneelmore:
___
*[container_service_extension/client/cse.py, line 513 at r1](https://reviewable.io/reviews/vmware/container-service-extension/350#-Lg9z669DXAZ1khuBSty:-LghoPMmC-8hdQcbEznN:b-42flm6) ([raw file](https://github.com/vmware/container-service-extension/blob/a3933bbe3e6d0fa660472505babe198d6e504590/container_service_extension/client/cse.py#L513)):*

How does this behave for sys admins, when they are using the system org (with no vdc selected)?

Btw I will suggest that we add the vdc and org flag for this command, and follow the logic we have used in other methods
```
        if vdc is None:
            vdc = ctx.obj['profiles'].get('vdc_in_use')
            if not vdc:
                raise Exception(f"Virtual datacenter context is not set. "
                                "Use either command 'vcd vdc use' or option "
                                "'--vdc' to set the vdc context.")
```

Think it from the perspective that we have to specify the org and vdc params in other commands to locate a cluster for resize, delete, etc. 
How would be able to target a cluster in this case from just the name of the cluster without provided the aforementioned params?
</blockquote></details>

Added the -org and -vdc flag for it to work for sys admin use case

harshneelmore:
___
*[container_service_extension/client/cse.py, line 513 at r1](https://reviewable.io/reviews/vmware/container-service-extension/350#-LgYEhDaAA53m5eeOeJ7:-Lh7yOF31WdK4P6-QBiY:b-4t7x0q) ([raw file](https://github.com/vmware/container-service-extension/blob/a3933bbe3e6d0fa660472505babe198d6e504590/container_service_extension/client/cse.py#L513)):*

We cannot assume vdc_in_use is the vdc of the cluster where our node is present.
For cluster info command, we do not pass 'vdc_in_use' by default.
If vdc is not passed as part of body, it is CSE server's responsibility to locate the cluster and its vdc. Similarly, For node info as well, it should be CSE server's respoinbility to identify the parent cluster's ovdc and then use the right broker to invoke the method.

Ideal fix for this:
Handle node_info in broker_manager.py
It's workflow should be similar to broker_manager._get_cluster_info()
</blockquote></details>

I have made the changes as directed, do take a look and let me know

harshneelmore:
___
*[container_service_extension/vcdbroker.py, line 289 at r3](https://reviewable.io/reviews/vmware/container-service-extension/350#-LhCeds7Ejv2lHU4IsAn:-LhCl-id5UlqncV8wCh5:b-47lqal) ([raw file](https://github.com/vmware/container-service-extension/blob/c06845f3c8307e4a5eb2197b41fc3e56a200ecf7/container_service_extension/vcdbroker.py#L289)):*

This change might have been checked in by Sompa or Sakthi. Do refresh your PR from the top of master branch.
</blockquote></details>

This is required as Node_info has to pass org and vdc options

harshneelmore:
___
*[container_service_extension/broker_manager.py, line 177 at r3](https://reviewable.io/reviews/vmware/container-service-extension/350#-LhCtDq433_pzXfJYtkS:-LhGx1N082v2m8D2mhBc:b-acmfo4) ([raw file](https://github.com/vmware/container-service-extension/blob/c06845f3c8307e4a5eb2197b41fc3e56a200ecf7/container_service_extension/broker_manager.py#L177)):*

What is the current behavior of "node list" command? Does it work as expected?
</blockquote></details>

Yes the current behavior stays in place

PR #358, opened 2019-06-10
Update init-photon-v2.sh

tironside:

- (Required) update init script to remove # from sshd_config AuthorizedKeysFile

- (Required) Current template builds wont allow ssh login as the authorized keys file location is commented out; this is a simple sed to fix it.

Signed-off-by: Tony Ironside <tony@ironside.ca>

### DIFF
--- a/scripts/init-photon-v2.sh
+++ b/scripts/init-photon-v2.sh
@@ -6,6 +6,8 @@ chage -I -1 -m 0 -M -1 -E -1 root
 
 ### Comment below line if you want to enable root password access
 ### root password access is dangerous; ssh-key access is recommended
+#AuthorizedKeysFile is commented out in /etc/ssh/sshd_config add in fix
+sed -i '/AuthorizedKeysFile/s/#//g' /etc/ssh/sshd_config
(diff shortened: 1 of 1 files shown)


andrew-ni:
Can you elaborate on why this change is needed? We have not had problems with ssh key access into VMs created using our current template scripts.

tironside:
To be honest it came about as part user error on a public key

it does actually work as is with a correct key, from a hygiene point of
view it is more consistent.

Thx

T

andrew-ni:
Would you be able to explain this in some more detail (why is this more hygienic?) and specify some use cases that require this fix?

Or if there's an external source where we can read more about this that'd be great too.

tironside:
Hi Andrew

from a hygene standpoint it is  a case of implied vs explicit parameters.
I don't like implied parameters (ssh keys etc) and prefer to have them
explicitly defined.

That being said; the ssh standard appears to be that this parameter is
optional unless you want to change it; I will cancel the pull, as it is
more a preference than a requirement at this point

Thanks!

T

PR #367, opened 2019-06-17
Change how admin passwords are generated and read for K8 cluster vms

rocknes:
While adding a node to a native K8 cluster, 
1. CSE assumes that the node vms admin password is always set.
2. CSE assumes the password is same as the one in the source template.

Both these assumptions can be wrong. The correct behavior in this case should be to auto generate a password while deploying the vm from template, read the password from the deployed VM instead of the template and update it to the user provided password.

Testing done:
* Used catalog external publish/subscribe model to create a copy of the CSE template in org1
* Manually checked that the subscribed catalog templates didn't have the vm password section.
* Deployed two K8 clusters (1 based on ubuntu template and 1 based on photon template) along with nfs servers, both deployments were successful.
* Also ran server and client system tests, both of them ran successfully.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/367)

### DIFF
--- a/container_service_extension/cluster.py
+++ b/container_service_extension/cluster.py
@@ -162,6 +162,7 @@ def add_nodes(qty, template, node_type, config, client, org, vdc, vapp, body):
                 'vapp': source_vapp.resource,
                 'target_vm_name': name,
                 'hostname': name,
+                'password_auto': True,
                 'network': body['network'],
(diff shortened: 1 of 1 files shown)


gmcelhoe:
:lgtm: too

PR #391, opened 2019-07-29
[VCDA-1105] Fix task based CSE operation for vCD 9.8

harshneelmore:
With vCD 9.8+ CSE is sending the wrong href_value while creating and updating tasks for long-running operations. This update will fix the href_value with cluster owner href_value and update owner type to 'application/vnd.vmware.vcloud.org+xml'.
- Manually tested p vCD 9.8 and 9.7

- @rocknes @andrew-ni @sakthisunda

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/391)

### DIFF
--- a/container_service_extension/vcdbroker.py
+++ b/container_service_extension/vcdbroker.py
@@ -149,6 +149,11 @@ class VcdBroker(AbstractBroker, threading.Thread):
         else:
             task_href = None
 
+        org_resource = self.tenant_client.get_org_by_name(
+            self.req_spec.get(RequestKey.ORG_NAME))
+        org = Org(self.tenant_client, resource=org_resource)
+        user_href = org.get_user(self.client_session.get('user')).get('href')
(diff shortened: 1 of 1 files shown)


harshneelmore:
___
*[container_service_extension/vcdbroker.py, line 152 at r1](https://reviewable.io/reviews/vmware/container-service-extension/391#-Ll-0qRE1_puvLRJmD1A:-Ll-Ay6TEklxVgweYk1q:b-et9osl) ([raw file](https://github.com/vmware/container-service-extension/blob/075efe0adfc65e1d5a10847ba0151fbccb947c52/container_service_extension/vcdbroker.py#L152)):*

This call will fail if it's a non sys admin user.

use the sys_admin client if you have access to it here.
</blockquote></details>

The call works with a non sys-admin user, just did the operation with vApp author.

PR #396, opened 2019-08-06
[VCDA-1093] Add persona based tests for cluster create and list

harshneelmore:
Added persona-based test support for cse create cluster and list clusters.
Additionally, refactored cse system enable/disable.
_ Tested the changes locally waiting tests run on Jenkins pipeline 

- @rocknes @andrew-ni @sakthisunda

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/396)

### DIFF
--- a/container_service_extension/system_test_framework/environment.py
+++ b/container_service_extension/system_test_framework/environment.py
@@ -58,7 +58,9 @@ SCRIPTS_DIR = 'scripts'
 
 SSH_KEY_FILEPATH = str(Path.home() / '.ssh' / 'id_rsa.pub')
 CLI_RUNNER = CliRunner()
-TEST_CLUSTER_NAME = 'testcluster'
+SYS_ADMIN_TEST_CLUSTER_NAME = 'testclustersystem'
+ORG_ADMIN_TEST_CLUSTER_NAME = 'testclusteradmin'
+VAPP_AUTHOR_TEST_CLUSTER_NAME = 'testclustervapp'
 
 # required user info
 ORG_ADMIN_NAME = 'orgadmin'
@@ -79,6 +81,7 @@ ORG_ADMIN_LOGIN_CMD = None
 VAPP_AUTHOR_LOGIN_CMD = None
 USER_LOGOUT_CMD = f"logout"
 USER_LOGIN_CMD_MAP = {}
+USERNAME_TO_CLUSTER_NAME = {}
 
 AMQP_USERNAME = None
 AMQP_PASSWORD = None
@@ -98,7 +101,8 @@ def init_environment(config_filepath=BASE_CONFIG_FILEPATH):
     global AMQP_USERNAME, AMQP_PASSWORD, CLIENT, ORG_HREF, VDC_HREF, \
         CATALOG_NAME, TEARDOWN_INSTALLATION, TEARDOWN_CLUSTERS, \
         TEMPLATE_DEFINITIONS, TEST_ALL_TEMPLATES, SYS_ADMIN_LOGIN_CMD, \
-        ORG_ADMIN_LOGIN_CMD, VAPP_AUTHOR_LOGIN_CMD, USER_LOGIN_CMD_MAP
+        ORG_ADMIN_LOGIN_CMD, VAPP_AUTHOR_LOGIN_CMD, USER_LOGIN_CMD_MAP, \
+        USERNAME_TO_CLUSTER_NAME
 
     config = testutils.yaml_to_dict(config_filepath)
 
@@ -141,6 +145,10 @@ def init_environment(config_filepath=BASE_CONFIG_FILEPATH):
     USER_LOGIN_CMD_MAP = {'sys_admin': SYS_ADMIN_LOGIN_CMD,
                           'org_admin': ORG_ADMIN_LOGIN_CMD,
                           'vapp_author': VAPP_AUTHOR_LOGIN_CMD}
+
+    USERNAME_TO_CLUSTER_NAME = {'sys_admin': SYS_ADMIN_TEST_CLUSTER_NAME,  # noqa
+                                         'org_admin': ORG_ADMIN_TEST_CLUSTER_NAME,  # noqa
+                                         'vapp_author': VAPP_AUTHOR_TEST_CLUSTER_NAME}  # noqa
 
     test_config = config.get('test')
     if test_config is not None:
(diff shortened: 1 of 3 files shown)


harshneelmore:
___
*[system_tests/test_cse_client.py, line 254 at r1](https://reviewable.io/reviews/vmware/container-service-extension/396#-Llhmb3a9otzxZ34eZC3:-Lli2E1S9Jkt2jhzcch7:b-de96x0) ([raw file](https://github.com/vmware/container-service-extension/blob/416e6244a888d50c474ef4331a4be8b533ba07db/system_tests/test_cse_client.py#L254)):*

Why is this fixture parameterized?

This is a post test cleanup function and should be run as *only* sys admin to cleanup the testbed.
</blockquote></details>

It's not just post-test clean up fixture it can be used to clean up the testbed of any exiting vApps before starting a new test

harshneelmore:
___
*[system_tests/test_cse_client.py, line 285 at r1](https://reviewable.io/reviews/vmware/container-service-extension/396#-LlhmwjIDLvFwLCuN6tN:-Lln1euMBm3cmxpl4jQu:b-896fix) ([raw file](https://github.com/vmware/container-service-extension/blob/416e6244a888d50c474ef4331a4be8b533ba07db/system_tests/test_cse_client.py#L285)):*

Why not just take the validator function as argument and if it's not None, then invoke it with the params?
We should avoid building this if else ladders in generic functions.
</blockquote></details>

Done.

harshneelmore:
___
*[system_tests/test_cse_client.py, line 307 at r1](https://reviewable.io/reviews/vmware/container-service-extension/396#-LlhnG326WfZOrPUefoY:-Lln1iX_DmSofWVmPMyk:b-896fix) ([raw file](https://github.com/vmware/container-service-extension/blob/416e6244a888d50c474ef4331a4be8b533ba07db/system_tests/test_cse_client.py#L307)):*

Add some comments about these numbers. How fragile are these asserts?
</blockquote></details>

Done.

harshneelmore:
___
*[system_tests/test_cse_client.py, line 254 at r1](https://reviewable.io/reviews/vmware/container-service-extension/396#-Llhmb3a9otzxZ34eZC3:-LlnC6KGC3G6cUUSgiC4:b-896fix) ([raw file](https://github.com/vmware/container-service-extension/blob/416e6244a888d50c474ef4331a4be8b533ba07db/system_tests/test_cse_client.py#L254)):*

This fixture should be run as only sys admin. 
The idea of this fixture is to do cleanup, right? In that case, why are different user personas required?

The delete cluster test which is a separate method altogether should have the parameterization.
</blockquote></details>

Done.

harshneelmore:
___
*[system_tests/test_cse_client.py, line 498 at r1](https://reviewable.io/reviews/vmware/container-service-extension/396#-Lli0Zm_0kw2pYPDCNIj:-Lln5Vwd0WIliMQg3Xba:b2t7ybe) ([raw file](https://github.com/vmware/container-service-extension/blob/416e6244a888d50c474ef4331a4be8b533ba07db/system_tests/test_cse_client.py#L498)):*

Might make sense to define constants for these, if we are using them at multiple places.
</blockquote></details>

Will make the changes in the next PR

harshneelmore:
___
*[system_tests/test_cse_client.py, line 267 at r2](https://reviewable.io/reviews/vmware/container-service-extension/396#-LlnMtGM6vkTcm3qJBWu:-Lls3AHh8E33f701fxjw:b-896fix) ([raw file](https://github.com/vmware/container-service-extension/blob/a88ee73110ab98c4934e3cfd56e3eb46b7b58ab8/system_tests/test_cse_client.py#L267)):*

Wrong indentation.

For loop deletes all three clusters then the method yields to caller, and on the way back if the flag is set run another loop to clean up the let over clusters.
</blockquote></details>

Done.

harshneelmore:
___
*[system_tests/test_cse_client.py, line 277 at r3](https://reviewable.io/reviews/vmware/container-service-extension/396#-Lm6qIFf8Y-eb50B-GT-:-LmGjjgf0sj16AkcLL83:b7vwbsf) ([raw file](https://github.com/vmware/container-service-extension/blob/307d1aef49b7c5f3f5476bf1c9af5a88a77a9ca1/system_tests/test_cse_client.py#L277)):*

these functions should not be in the `test_x.py` files, as these files should only contain fixtures and actual test functions. Can these miscellaneous functions be moved to environment.py or system_test_framework/utils.py?
</blockquote></details>

Moving it to system_tests/utils.py makes more sense there

harshneelmore:
___
*[system_tests/test_cse_client.py, line 627 at r3](https://reviewable.io/reviews/vmware/container-service-extension/396#-Lm6vgGd4w30TiYcs1iA:-LmGtCkL1zpr9jYX3YTU:b-896fix) ([raw file](https://github.com/vmware/container-service-extension/blob/307d1aef49b7c5f3f5476bf1c9af5a88a77a9ca1/system_tests/test_cse_client.py#L627)):*

Adding onto my point above, it seems like we're breaking test independence. If we run this test by itself, there won't be any clusters to list
</blockquote></details>

Done.

harshneelmore:
___
*[system_tests/test_cse_client.py, line 587 at r3](https://reviewable.io/reviews/vmware/container-service-extension/396#-Lm6ub3V1bZ4V5m1ie2j:-LmGtBsR3Adw7gah2NdQ:b-896fix) ([raw file](https://github.com/vmware/container-service-extension/blob/307d1aef49b7c5f3f5476bf1c9af5a88a77a9ca1/system_tests/test_cse_client.py#L587)):*

Just a note: if we're splitting up all cluster operations to separate tests, we need to keep in mind that all tests should be independent of all other tests, so in this model, `test_delete_cluster` should be able to be run by itself.

The reasoning behind the big `test_cluster_operations` function was to reduce cluster deployment time and only allow testing the cluster lifecycle as a whole, as most of the time, we'd want to test most/all of the cluster operations together anyways.
</blockquote></details>

Done.

harshneelmore:
___
*[system_tests/test_cse_client.py, line 562 at r3](https://reviewable.io/reviews/vmware/container-service-extension/396#-Lm6tuNfBeuHWGZO4kIx:-LmLPwvMAHdqcOSCOrVA:b-f8lm0k) ([raw file](https://github.com/vmware/container-service-extension/blob/307d1aef49b7c5f3f5476bf1c9af5a88a77a9ca1/system_tests/test_cse_client.py#L562)):*

this doesn't seem like it should be a test, but rather a fixture probably? Because this isn't testing CSE. This can probably be done in the CSE server module level fixture in this module.
</blockquote></details>

Moved it inside the server module fixture.

PR #402, opened 2019-08-15
[VCDA-1094] Add persona support for delete clusters

harshneelmore:
Add support to delete clusters by using different personas - Vapp Author, Org admin. 
This test will delete the cluster created by cse create cluster command.
Tested manually waiting to test on Jenkins

@rocknes @andrew-ni @sakthisunda

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/402)

### DIFF
--- a/container_service_extension/system_test_framework/environment.py
+++ b/container_service_extension/system_test_framework/environment.py
@@ -63,10 +63,11 @@ ORG_ADMIN_TEST_CLUSTER_NAME = 'testclusteradmin'
 VAPP_AUTHOR_TEST_CLUSTER_NAME = 'testclustervapp'
 
 # required user info
-ORG_ADMIN_NAME = 'orgadmin'
+SYS_ADMIN_NAME = 'sys_admin'
+ORG_ADMIN_NAME = 'org_admin'
 ORG_ADMIN_PASSWORD = 'password'
 ORG_ADMIN_ROLE_NAME = 'Organization Administrator'
-VAPP_AUTHOR_NAME = 'vappauthor'
+VAPP_AUTHOR_NAME = 'vapp_author'
 VAPP_AUTHOR_PASSWORD = 'password'
 VAPP_AUTHOR_ROLE_NAME = 'vApp Author'
 
--- a/system_tests/test_cse_client.py
+++ b/system_tests/test_cse_client.py
@@ -58,7 +58,7 @@ import time
 import pytest
 from vcd_cli.vcd import vcd
 
-from container_service_extension.cse import cli
+from container_service_extension.server_cli import cli
 import container_service_extension.server_constants as constants
 import container_service_extension.system_test_framework.environment as env
 import container_service_extension.system_test_framework.utils as testutils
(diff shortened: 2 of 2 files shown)


harshneelmore:
___
*[container_service_extension/system_test_framework/environment.py, line 80 at r1](https://reviewable.io/reviews/vmware/container-service-extension/402#-LmN-2Ff5n2oPVIK1Z7n:-LmQkF7tC_XE9seIjDVj:b-37tgyi) ([raw file](https://github.com/vmware/container-service-extension/blob/da4d6059acf34b0d3125f265d0be648670afed66/container_service_extension/system_test_framework/environment.py#L80)):*

Just curious, are these values intended as the sentinel values for test parameterization?

In either case, I agree with aritra in that these are already defined in the above `# required user info` section. If we are using sentinel values for test parameterization, then we shouldn't create these new constants just for that when we can re-use the constants already defined. (We can use `ORG_ADMIN_NAME` and `VAPP_AUTHOR_NAME` AND `SYS_ADMIN_NAME` for the sentinel values)
</blockquote></details>

I made new constants for the #systemtests as there were existing tests using the old names and decided to create new ones for now and then when I refactor and clean up the tests to remove the old tests I could make changes to the constants and remove any redundant names.

harshneelmore:
___
*[system_tests/test_cse_client.py, line 468 at r2](https://reviewable.io/reviews/vmware/container-service-extension/402#-LmQiKbA-jYFTi5YrT5_:-LmQkijq2CziNL7dYNeP:b-x49fla) ([raw file](https://github.com/vmware/container-service-extension/blob/5e87daed58cd92d019b7158f6abb90e71be60fa4/system_tests/test_cse_client.py#L468)):*

these constant strings for `test_runner_username` should all be replaced with the constants defined in environment.py (SYS_ADMIN_NAME, VAPP_AUTHOR_NAME, ORG_ADMIN_NAME)

same comment for all other tests parameterized this way
</blockquote></details>

I had left it for later refactoring and cleaning up the code. But I went ahead made the changes in this PR

harshneelmore:
___
*[system_tests/test_cse_client.py, line 629 at r1](https://reviewable.io/reviews/vmware/container-service-extension/402#-LmQfnpA2YEwcKCMXLm1:-LmQkvqnAt1vngGxoArL:b-ipop4e) ([raw file](https://github.com/vmware/container-service-extension/blob/da4d6059acf34b0d3125f265d0be648670afed66/system_tests/test_cse_client.py#L629)):*

when using these constant mappings that are defined by us, we should avoid using `.get()` and use `[]`. Reason being is that if any of these constants change in the future, we don't attempt this command with a `None` value, but rather we immediately fail with a KeyError, which is much easier to debug.

`.get()` is better left for when we don't expect to fail on receiving `None`
</blockquote></details>

Agreed next time the constants are changed we get a failure KeyError instead of None.

harshneelmore:
___
*[system_tests/test_cse_client.py, line 641 at r1](https://reviewable.io/reviews/vmware/container-service-extension/402#-LmN-alC3Sy8endhzyVn:-LmQl9iJAFA8w3xULI6Q:b-896fix) ([raw file](https://github.com/vmware/container-service-extension/blob/da4d6059acf34b0d3125f265d0be648670afed66/system_tests/test_cse_client.py#L641)):*

```env.vapp_exists(env.USERNAME_TO_CLUSTER_NAME.get(cluster))```

I think this is not correct. Your for loop is iterating through cluster names already. Passing the cluster_name into the constant map `USERNAME_TO_CLUSTER_NAME` will always give you `None`, because you aren't providing any USERNAME.

This should be rewritten as:
```
for cluster_name in env.USERNAME_TO_CLUSTER_NAME.values():
    assert not env.vapp_exists(cluster_name), "Cluster exists when it should not"
```
</blockquote></details>

Done.

harshneelmore:
___
*[system_tests/test_cse_client.py, line 285 at r3](https://reviewable.io/reviews/vmware/container-service-extension/402#-LmQmfYv4XpdIWSXWA_B:-LmQnZuGCcQGVqjLBLWT:b-allill) ([raw file](https://github.com/vmware/container-service-extension/blob/2e573c43f82d3b549ae54c9d4f54c453141f0394/system_tests/test_cse_client.py#L285)):*

was this second vapp delete for vapp author intentional?
</blockquote></details>

Not sure how it got an extra one, removing it

PR #429, opened 2019-09-16
[VCDA-1195] Force compute policy removal from template to be done only by API 32.0

harshneelmore:
vCD API v33.0 has a bug in it, due to which the call to remove compute policy from a template VM becomes a no-op. This fix will invoke the corresponding method in pyvcloud, to remove compute policy using a VCD API client version 32.0.
- Tested successfully on vCD 10.

- @rocknes @sakthisunda @andrew-ni

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/429)

### DIFF
--- a/container_service_extension/compute_policy_manager.py
+++ b/container_service_extension/compute_policy_manager.py
@@ -2,6 +2,7 @@
 # Copyright (c) 2019 VMware, Inc. All Rights Reserved.
 # SPDX-License-Identifier: BSD-2-Clause
 
+from pyvcloud.vcd.client import ApiVersion
 from pyvcloud.vcd.client import find_link
 from pyvcloud.vcd.exceptions import MissingLinkException
 from pyvcloud.vcd.exceptions import OperationNotSupportedException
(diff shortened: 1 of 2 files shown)


harshneelmore:
___
*[container_service_extension/compute_policy_manager.py, line 282 at r1](https://reviewable.io/reviews/vmware/container-service-extension/429#-Lp00lPJ7JYcXwaVajXd:-Lp0BF1VBJBk3cwB1S-w:b-896fix) ([raw file](https://github.com/vmware/container-service-extension/blob/ff6c6d1583f674253bd294f9850e24e617bc8015/container_service_extension/compute_policy_manager.py#L282)):*

We can get rid of this version check if we follow my suggestion in the previous comment, and straight up use 

self._vcd_client_hack 
</blockquote></details>

Done.

rocknes:
We have identified a different approach to solve the problem. Issue addressed in https://github.com/vmware/container-service-extension/pull/435
Closing this PR.

PR #461, opened 2019-10-07
Fix bug blocking new compute policy creation and server startup

andrew-ni:
An instance of get_policy() raising an exception was not
being handled. This prevents CSE from creating new
compute policies and starting up the server when a new
compute policy is specified in template rules

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/461)

### DIFF
--- a/container_service_extension/compute_policy_manager.py
+++ b/container_service_extension/compute_policy_manager.py
@@ -112,8 +112,9 @@ class ComputePolicyManager:
 
         :param str policy_name: name of the compute policy
 
-        :return: policy details if found, else None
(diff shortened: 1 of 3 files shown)


goelaashima:
*[a discussion](https://reviewable.io/reviews/vmware/container-service-extension/461#-Lql8tVK8gUPTmXRcIyO:-Lql8tVK8gUPTmXRcIyP:bvzik0r) (no related file):*
Andrew - Can you move release notes into a separate pull request? Also remove root-cause section and fixed-in section. Doc will come right around patch.

---

PR #557, opened 2020-04-03
[VCDA-1451] Update cluster list to show Kubernetes info

andrew-ni:
- Fixed bug in pksbroker._get_token() error handling where `requests.code` was used instead of `requests.codes`. This issue would appear if the user's CSE-PKS integration was set up incorrectly

- Updated cluster list API to send back 'k8s_type' data

- Updated CLI cluster list to display cluster's Kubernetes and display more user friendly headers

Before this change:
image

After this change:
image

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/557)


Anirudh9794:
:lgtm:

PR #558, opened 2020-04-03
[VCDA-1431] resolve multiple telemetry calls

Anirudh9794:

- - Avoid multiple calls to record telemetry details. Happens because the following functions which are supposed to be only by the request_processor are used in other functions as well:
  - get_cluster_info
  - create_nodes
  - get_cluster_upgrade_plan

- broker functions will now accept kwargs which can have data along with telemetry flag.
- record telemetry details only if telemetry flag is set

- testing done:
added debug statements to check if multiple calls to record telemetry are being made
`vcd cse cluster resize -N 3 -n mynet clus3`
`vcd cse cluster list`
`vcd cse cluster info clus3`
`vcd cse node info clus3 mstr-xkfv`

@rocknes @andrew-ni @sakthisunda @sahithi

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/558)

### DIFF
--- a/container_service_extension/abstract_broker.py
+++ b/container_service_extension/abstract_broker.py
@@ -30,7 +30,7 @@ class AbstractBroker(abc.ABC):
             self.tenant_client._get_wk_endpoint(_WellKnownEndpoint.LOGGED_IN_ORG) # noqa: E501
 
     @abc.abstractmethod
-    def create_cluster(self, data):
+    def create_cluster(self, **kwargs):
         """Create cluster.
 
         :return: response object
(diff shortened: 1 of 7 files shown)


Anirudh9794:
___
*[container_service_extension/broker_manager.py, line 61 at r2](https://reviewable.io/reviews/vmware/container-service-extension/558#-M4tfPnT0PEpqNBq7fHJ:-M4u2wNZ7yOeYT197lNp:b-896fix) ([raw file](https://github.com/vmware/container-service-extension/blob/004feb151bebfc5f8241e03a52643fdc39285fad/container_service_extension/broker_manager.py#L61)):*

Just pass the kwargs as is, the broker will pick up the 'telemetry' flag if it has to.

We should not add the telemetry param in every method signature.
</blockquote></details>

Done.

PR #571, opened 2020-04-24
[VCDA-1477] Implement request context

andrew-ni:
Write up: https://confluence.eng.vmware.com/pages/viewpage.action?pageId=614902423

- RequestContext is an abstraction for vcd api clients, and user-specific details.
- This abstracts away details involving user/sysadmin client creation/logout, obtaining user details (username, role, rights, id)

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/571)

### DIFF
--- a/container_service_extension/abstract_broker.py
+++ b/container_service_extension/abstract_broker.py
@@ -4,30 +4,12 @@
 
 import abc
 
-from pyvcloud.vcd.client import _WellKnownEndpoint
-
-import container_service_extension.pyvcloud_utils as vcd_utils
+import container_service_extension.request_context as ctx
 
 
 class AbstractBroker(abc.ABC):
-
-    def __init__(self, tenant_auth_token, is_jwt_token):
-        self.tenant_client = None
-        self.client_session = None
-        self.tenant_user_name = None
-        self.tenant_user_id = None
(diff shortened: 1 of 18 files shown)


Anirudh9794:
___
*[container_service_extension/broker_manager.py, line 58 at r1](https://reviewable.io/reviews/vmware/container-service-extension/571#-M5hnDCo2zh8Mui8w1i1:-M5hzYwm87SVCbnpp-0O:bbpvzi8) ([raw file](https://github.com/vmware/container-service-extension/blob/3618671cbe67485c182300d860234d93122ca536/container_service_extension/broker_manager.py#L58)):*

by runtime configuration do you mean the server config dictionary?

I was thinking about including the request_data into the request_context, but thought against it for this PR at least, because the intended use for request_data was that the handlers would pick out relevant data and pass them into any helper functions instead of always passing around an ambiguous data dictionary. For example ovdc_handler.ovdc_info() follows this, but our brokers are an exception to this. We can probably discuss this over a call, but if we do decide the bundle the data with context, i'd like to do that in a separate PR.


</blockquote></details>

I was mentioning server config dictionary because it is used many times inside the handlers and also contains decision variables (like log_wire) but it is also okay if it is fetched from Service.

PR #588, opened 2020-05-07
VCDA-1498 Defined entity schema service

sahithi:
1. Implemented **schema_svc.py** provides API to perfom CRUD on _defined entity interfaces_ and _entity types_.
2. **models.py** has local model classes for _Defined entity interfaces_ and _entity types_.
3. Created a placeholder file **entity_svc.py** to handle CRUD on _entities._

**Notes to reviewers:**
I will modify **schema_svc.py** to pull cloud-api client from request-context once that change is done.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/588)


Anirudh9794:
___
*[container_service_extension/def_modules/schema_svc.py, line 65 at r2](https://reviewable.io/reviews/vmware/container-service-extension/588#-M77AaDu3S0N8lzTipjt:-M79wCSB7Ww5n8cDeXFM:bjngmla) ([raw file](https://github.com/vmware/container-service-extension/blob/e8f750e413748529ae400ff9128a58bb920278ef/container_service_extension/def_modules/schema_svc.py#L65)):*

Aritra, as discussed, I copied this init block from ComputePolicyManager. In fact, I did wonder about the need for this logic. The real check that needs to happen is - if vCD API version is >= X, why make a call on a wire for this?

I completely removed this logic and add a TODO to do an API version check at right place when we exactly identify the location for it.
@anirudh9794 Can you make the similar change in ComputePolicyManager?
</blockquote></details>

Will make the change. Thanks!

PR #618, opened 2020-06-16
Workaround for GC bug in vCD leading to template creation failure

rocknes:
vCD recently consumed VC 7.0 imgcust build. The new imgcust build has a severe bug in it, and as a result guest customization(GC) is in broken in vCD. CSE uses GC to create k8s templates as well k8s clusters. This bug impacts CSE directly and hampers most basic workflows in CSE. To counter this issue, this PR removes all usage of GC in CSE and replaces them with vanilla script execution in vm via vCenter.

Testing done:
Created kubernetes 1.17 template - result : template creation successful
Created a cluster off the same template - result : cluster creation successful

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/618)

### DIFF
--- a/container_service_extension/template_builder.py
+++ b/container_service_extension/template_builder.py
@@ -202,8 +202,6 @@ class TemplateBuilder():
         except EntityNotFoundException:
             pass
 
-        init_script = self._get_init_script()
-
         msg = f"Creating vApp '{self.temp_vapp_name}'"
         self.msg_update_callback.info(msg)
         self.logger.info(msg)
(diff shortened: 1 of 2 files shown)


rocknes:
vCD recently consumed VC 7.0 imgcust build. The new imgcust build has a severe bug in it, and as a result guest customization(GC) is in broken in vCD. CSE uses GC to create k8s templates as well k8s clusters. This bug impacts CSE directly and hampers most basic workflows in CSE. To counter this issue, this PR removes all usage of GC in CSE and replaces them with vanilla script execution in vm via vCenter.

PR #641, opened 2020-07-07
[VCDA-1580] Move to pre-defined kubernetes interface

Anirudh9794:

Make use of built in kubernetes interface for creating native cluster defined entity type

Testing done:
 `cse upgrade -c config -st`
checked to see if kubernetes interface was loaded on `cse run`

@sahithi @sakthisunda @rocknes @andrew-ni

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/641)

### DIFF
--- a/container_service_extension/configure_cse.py
+++ b/container_service_extension/configure_cse.py
@@ -1,6 +1,8 @@
 # container-service-extension
 # Copyright (c) 2017 VMware, Inc. All Rights Reserved.
 # SPDX-License-Identifier: BSD-2-Clause
+import importlib
+import importlib.resources as pkg_resources
 import json
 
 import pika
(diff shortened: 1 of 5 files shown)


Anirudh9794:
___
*[container_service_extension/service.py, line 388 at r1](https://reviewable.io/reviews/vmware/container-service-extension/641#-MBeQdJmB0J5EXYboQ1Y:-MBeVIQv9L9TjHIzNPVT:b9jvw2y) ([raw file](https://github.com/vmware/container-service-extension/blob/83d0d6f927a1c892f3dd22803ee17d0383363bd3/container_service_extension/service.py#L388)):*

Curious why only interface_id uses defKey.VMWARE_VENDOR but entity_type_id uses defKey.VENDOR still? 

Or is there no relation betweeen these two? 
</blockquote></details>

I think there is no relation between these two. Just that the interface is created by vcd and the entity type is created by cse. I discussed this with Sahithi and she said this is okay.

PR #649, opened 2020-07-13
[VCDA-1624] Hook TKG client to CSE client

sahithi:
Added Tkg client bindings into CSE client.
As there were no GET and LIST API by default, I had to write those methods in their _tkg_cluster_api.py_ to leverage model classes.

**Changed files**
_def_entity_cluster.py_ - Sample GET and LIST methods for TKG clusters. They can be modified later to cater to our command outputs.
_api_client.py_ -  Tweaked _deserialization()_ a bit to accommodate the lack of model class for defined entity.
_tkg_cluster_api.py_ - API calls for GET and LIST - _get_tkg_cluster()_ and _list_tkg_cluster()_

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/649)


Anirudh9794:
___
*[container_service_extension/client/tkgclient/api/tkg_cluster_api.py, line 67 at r2](https://reviewable.io/reviews/vmware/container-service-extension/649#-MCE7hi37qSfuoJ6Z8TE:-MCE7hi37qSfuoJ6Z8TF:b-w03mad) ([raw file](https://github.com/vmware/container-service-extension/blob/fb84e90f1087ec599f03d00bdae5faa61a795d22/container_service_extension/client/tkgclient/api/tkg_cluster_api.py#L67)):*
> ```Python
>         auth_settings = []
>         collection_formats = {}
>         return self.api_client.call_api(
> ```

call api will make a single request. Either path_params should take in paging parameters or a generator should be returned.

Anirudh9794:
___
*[container_service_extension/client/tkgclient/api/tkg_cluster_api.py, line 83 at r2](https://reviewable.io/reviews/vmware/container-service-extension/649#-MCF2nWO0OIzFEBS0BJ8:-MCF2nWO0OIzFEBS0BJ9:b-ou70dx) ([raw file](https://github.com/vmware/container-service-extension/blob/fb84e90f1087ec599f03d00bdae5faa61a795d22/container_service_extension/client/tkgclient/api/tkg_cluster_api.py#L83)):*
> ```Python
>             collection_formats=collection_formats)
> 
>     def get_tkg_cluster(self, id, **kwargs):
> ```

This method is very close to entity_service.get_entity(). Can the method in entity_service be modified instead of adding an additional method? I think that the problem would be multiple types of output. But will help avoid making if checks every time a get_entity() is called. I think this is something that can be explored later

PR #672, opened 2020-07-28
[VCDA-1517] Ignore cse node commands for api version 35

Anirudh9794:

- Ignore `vcd cse node ...` commands for api version >= 35

- Created command level filter for cse and ignored the  `node` command for apiVersion >= 35

Sample output:
Screen Shot 2020-07-27 at 6 01 32 PM

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/672)

### DIFF
--- a/container_service_extension/client/command_filter.py
+++ b/container_service_extension/client/command_filter.py
@@ -32,6 +32,12 @@ class CommandNameKey(str, Enum):
 
 # List of unsupported commands by Api Version
 UNSUPPORTED_COMMANDS_BY_VERSION = {
+    vcd_client.ApiVersion.VERSION_35.value: [GroupKey.NODE]
+}
+
+
+# List of unsupported subcommands by Api Version
+UNSUPPORTED_SUBCOMMANDS_BY_VERSION = {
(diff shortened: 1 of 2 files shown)


Anirudh9794:
___
*[container_service_extension/client/cse.py, line 32 at r1](https://reviewable.io/reviews/vmware/container-service-extension/672#-MDLUKKK1DhxfyhHH6ob:-MDRM_vs9TOBEmgAROQQ:bi8e9dr) ([raw file](https://github.com/vmware/container-service-extension/blob/37f861adb057a04aa2c860544d69500b3d9a8ce0/container_service_extension/client/cse.py#L32)):*

Since GroupCommandFilter is introduced at @vcd.group and already existing @cse.group level, please check if there is any side effect like multiple calls to get_command().
</blockquote></details>

The filter needs to be present at @vcd.group and also at the @cse.group level. `get_command()` will be called twice - First call to filter out command and second time to filter out sub commands. I don't think it should be a problem unless the command and subcommand have the same name. (One use case I can think of is in `vcd cse pks ...` commands)

PR #680, opened 2020-07-31
[VCDA-1675, VCDA-1630] config validator and install with mqtt components

ltimothy7:
Signed-off-by: Timothy Le <ltimothy@vmware.com>

- (Required) Short description of changes in the PR topic line
Added mqtt handling to the config validator, cse install, and checking the cse installation.

The config validator, cse installation, and checking cse installation logic now include installing and checking the mqtt components if a config file has an "mqtt" section and if the api version is greater than or equal to 35.0. Also, the sample config in the config validator also includes the sample mqtt field(s) when mqtt is being used. 

The system tests have run successfully on ctot for 10.1, and most of the tests have passed for 10.2; I reviewed the error for ctot 10.2 tests, and it appears to possibly from not having the latest official build for vcd.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/680)


container_service_extension/mqtt_extension_manager.py
```diff
@@ -398,6 +402,7 @@ def delete_extension_token(self, token_id):
         """Delete the specified token with token_id.
 
         :param str token_id: the token's id
+
```
ChintanShahVMware, line 405: Whitespace change... not sure if intentional.

container_service_extension/mqtt_extension_manager.py
```diff
@@ -534,6 +539,7 @@ def delete_api_filter(self, filter_id):
         """Delete the api filter.
 
         :param str filter_id: filter id
+
```
ChintanShahVMware, line 542: Another whitespace change...

container_service_extension/sample_generator.py
```diff
@@ -72,6 +72,12 @@
     }
 }
 
+SAMPLE_MQTT_CONFIG = {
+    'mqtt': {
```
ChintanShahVMware, line 76: interesting, is there bare minimum config necessary for MQTT? 
So unlike AMQP, we don't need to configure sample host/port/username/password etc?
I am unfamiliar with MQTT but this seems a little confusing to me...
Please double check, and in case its actually not necessary, please add a comment describing min sample config for MQTT in here.
That will educate people like me :)

container_service_extension/utils.py
```diff
@@ -365,3 +367,19 @@ def wrapper(*args, **kwargs):
 def generate_thread_name(function_name):
     parent_thread_id = threading.current_thread().ident
     return function_name + ':' + str(parent_thread_id)
+
+
+def use_mqtt_protocol(config):
```
ChintanShahVMware, line 372: Probably this is not directly related to : 
use_mqtt_protocol function;
but what will CSE do in case both AMQP/MQTT config sections are present in config.yaml?
Is there a code making sure such confusing configs are invalidated? Do we need to add such check somewhere?

Tagging Aritra @rocknes in case we need some more clarity/explanation here.

system_tests/base_config.yaml
```diff
@@ -68,4 +68,4 @@ broker:
   org: '???'                            # vCD org that contains the shared catalog where the master templates will be stored
   remote_template_cookbook_url: https://raw.githubusercontent.com/vmware/container-service-extension-templates/upgrades/template.yaml
   storage_profile: '*'                  # name of the storage profile to use when creating the temporary vApp used to build the template
-  vdc: '???'                            # VDC within @org that will be used during the install process to build the template
+  vdc: '???'                            # VDC within @org that will be used during the install process to build the template
```
ChintanShahVMware, line 71: whitespace change?
Please remove it.

container_service_extension/configure_cse.py
```diff
@@ -1162,7 +1203,7 @@ def _upgrade_to_35(client, config, ext_vcd_api_version,
                    admin_password, msg_update_callback=utils.NullPrinter(),
                    log_wire=False):
     # Update amqp exchange
-    _create_amqp_exchange(
+    _create_amqp_exchange(  # TODO: check if using mqtt
```
ChintanShahVMware, line 1206: So this TODO Will get cleaned up as part of same release or later?
Lets have offline discussion on this :)

container_service_extension/configure_cse.py
```diff
@@ -296,20 +321,50 @@ def install_cse(config_file_name, skip_template_creation,
...
+        else:
+            # create amqp exchange if it doesn't exist
+            amqp = config['amqp']
+            _create_amqp_exchange(amqp['exchange'], amqp['host'], amqp['port'],
+                                  amqp['vhost'], amqp['ssl'], amqp['username'],
+                                  amqp['password'],
+                                  msg_update_callback=msg_update_callback)
+
+            # register cse as an api extension to vCD
+            _register_cse_as_extension(  # TODO: see how to use MQTT
```
ChintanShahVMware, line 347: Its slightly confusing; So we should:

- Registering CSE as an extension

- Register rights

only if the message-bus is AMQP? in case of MQTT we do not need to do that?
Can we please add some clarity about it in design doc, if thats actually the case?

container_service_extension/configure_cse.py
```diff
@@ -121,47 +145,45 @@ def check_cse_installation(config, msg_update_callback=utils.NullPrinter()):
                 msg_update_callback.error(msg)
                 SERVER_CLI_LOGGER.error(msg)
                 err_msgs.append(msg)
-        except Exception:  # TODO() replace raw exception with specific
```
ChintanShahVMware, line 124: Not sure why do we need to remove catching generic Exception.
Although I am not a fan of catching generic Exceptions; would like to understand the reasoning here.

container_service_extension/configure_cse.py
```diff
@@ -95,19 +97,41 @@ def check_cse_installation(config, msg_update_callback=utils.NullPrinter()):
...
-                                               ssl=amqp['ssl'],
-                                               connection_attempts=3,
-                                               retry_delay=2, socket_timeout=5)
-        connection = None
-        try:
-            connection = pika.BlockingConnection(parameters)
-            channel = connection.channel()
+        if utils.use_mqtt_protocol(config):
+            # Check that MQTT extension exists
+            mqtt_ext_manager = MQTTExtensionManager(client)
```
ChintanShahVMware, line 102: Can break down this code into functions for AMQP/MQTT specific operations?
i.e.

`
if utils.use_mqtt_protocol(config):
    do_things_specific_to_mqtt(ARGS)
`
`
else:
    do_things_specific_to_amqp(ARGS)
`
This way it will be more readable?

Right now the methods like check_cse_installation and install_cse are getting way to big.
Its hard to follow all the operations in long functions.

container_service_extension/configure_cse.py
```diff
@@ -121,47 +145,45 @@ def check_cse_installation(config, msg_update_callback=utils.NullPrinter()):
...
-                    msg += f"\nvCD-CSE exchange: {cse_info['exchange']}" \
-                           f"\nCSE config exchange: {amqp['exchange']}"
-                msg_update_callback.info(msg)
-                SERVER_CLI_LOGGER.info(msg)
+            # check that CSE is registered to vCD correctly
+            ext = api_extension.APIExtension(client)
+            try:
+                cse_info = ext.get_extension(server_constants.CSE_SERVICE_NAME,
+                                             namespace=server_constants.CSE_SERVICE_NAMESPACE) # noqa: E501
+                rkey_matches = cse_info['routingKey'] == amqp['routing_key']
```
ChintanShahVMware, line 157: Pardon my ignorance, but something doesn;t look right here.
We added a check above in the same function to see if we need to configure MQTT or AMQP.
Now at this stage are we guaranteed to have AMQP configs?
i.e. amqp['routing_key'], amqp['exchange'] etc?

container_service_extension/config_validator.py
```diff
@@ -103,8 +105,11 @@ def get_validated_config(config_file_name,
     msg_update_callback.info(
         f"Validating config file '{config_file_name}'")
     # This allows us to compare top-level config keys and value types
+    use_mqtt = use_mqtt_protocol(config)
+    sample_protocol_dict = SAMPLE_AMQP_CONFIG if not use_mqtt \
+        else SAMPLE_MQTT_CONFIG
     sample_config = {
-        **SAMPLE_AMQP_CONFIG, **SAMPLE_VCD_CONFIG,
+        **sample_protocol_dict, **SAMPLE_VCD_CONFIG,
```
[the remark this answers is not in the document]
ltimothy7: Hi Chintan, I agree that your renaming of "sample_protocol_dict" is more consistent and am changing it. I am leaving it lower cased because "sample_protocol_dict" is a local variable that is either the SAMPLE_AMQP_CONFIG or the SAMPLE_MQTT_CONFIG.

container_service_extension/config_validator.py
```diff
@@ -115,7 +120,8 @@ def get_validated_config(config_file_name,
         nsxt_wire_logger = SERVER_NSXT_WIRE_LOGGER
     check_keys_and_value_types(config, sample_config, location='config file',
                                msg_update_callback=msg_update_callback)
-    _validate_amqp_config(config['amqp'], msg_update_callback)
+    if not use_mqtt:
+        _validate_amqp_config(config['amqp'], msg_update_callback)
```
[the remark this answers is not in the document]
ltimothy7: From what I understand, we do not need to validate the mqtt config components because the only field we currently have for the mqtt section is a "verify_ssl" flag. In contrast, for amqp, the host, username, and password must be specified in the config file and validated. For the mqtt extension, we register it ourselves and do not currently need a host or other information. (You could look at the install_cse function in container_service_extension/configure_cse.py to see how we install the mqtt extension without additional information from the config.) Thank you for checking!

container_service_extension/mqtt_extension_manager.py
ltimothy7: Thank you, Chintan. I just added it to try to be consistent with container_service_extension/compute_policy_manager.py, which separate the :param section from the :raise/:return section of the comments.

container_service_extension/mqtt_extension_manager.py
ltimothy7: Same reason as above :)

container_service_extension/sample_generator.py
ltimothy7: For mqtt, we register the extension directly using the cloudapi and api clients, so we don't need information such as the hose, port, etc. We need the verify_ssl flag to see whether or not to enable security settings. For example, when testing with a host with a well-signed certificate, verify_ssl can be set to true. To see how we create the extension and its components, please see container_service_extension/mqtt_extension_manager.py and look for the create_* functions (for creating the extension, token, and api filter)

container_service_extension/utils.py
ltimothy7: As of now, if there is the presence of the mqtt section in config.yaml and the vcd api version is greater than or equal to the minimum required version for MQTT, then mqtt will be used even in the presence of an amqp section. Otherwise, amqp will be used. I agree we could logic to make sure only one of these fields is present would be helpful too. @rocknes What do you think?

container_service_extension/configure_cse.py
ltimothy7: Thank you, I have changed this.

container_service_extension/configure_cse.py
ltimothy7: This is a helpful point. I think this touches on your earlier comment we tagged @rocknes in. We are currently assuming that amqp is the default, and mqtt is optional to use instead if the user specifies.

 In get_validated_config in container_service_extension/config_validator.py, this code makes sure that the config will either have an "mqtt" or "amqp" section, depending on the the use_mqtt_protocol helper in utils. Therefore, at this point in the code, if MQTT is not used, there is an AMQP section. This is a roundabout reasoning, but we have the guarantee from the config validator that we will have either the amqp or mqtt sections; otherwise, we would have had an error earlier.

PR #690, opened 2020-08-11
[VCDA-1704] Default down to tkg only operations when cse server is not running

Anirudh9794:

Decisions made:
* When run for the first time, the cse cli tries to contact cse server. If the server is not running, it will set the flag `cse_server_running` in profiles.yaml
* Subsequent calls will first check two keys `cse_server_running` and `cse_server_api_version` in the profiles yaml
* If CSE server is not running, all the cli calls will land to `tkg_cluster_api.yaml`
* If the CSE server starts running after the flag `cse_server_running` is set, the user has to log out

Screenshot:
Screen Shot 2020-08-11 at 2 49 08 PM

@sahithi @rocknes @sakthisunda

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/690)

### DIFF
--- a/container_service_extension/client/utils.py
+++ b/container_service_extension/client/utils.py
@@ -7,8 +7,26 @@ import requests
 from vcd_cli.profiles import Profiles
 
 from container_service_extension.client import system as syst
+from container_service_extension.client.constants import CSE_SERVER_API_VERSION
+from container_service_extension.client.constants import CSE_SERVER_RUNNING
 import container_service_extension.def_.utils as def_utils
-from container_service_extension.shared_constants import CSE_SERVER_API_VERSION
+
+_ENABLE_ONLY_TKG_OPERATIONS = False
+
+
+def is_cli_for_tkg_only():
+    global _ENABLE_ONLY_TKG_OPERATIONS
+    return _ENABLE_ONLY_TKG_OPERATIONS
+
+
+def enable_cli_for_only_tkg_operations():
+    global _ENABLE_ONLY_TKG_OPERATIONS
+    _ENABLE_ONLY_TKG_OPERATIONS = True
+
+
+def disable_cli_for_only_tkg_operations():
+    global _ENABLE_ONLY_TKG_OPERATIONS
+    _ENABLE_ONLY_TKG_OPERATIONS = False
 
 
 def cse_restore_session(ctx, vdc_required=False) -> None:
@@ -49,7 +67,7 @@ def cse_restore_session(ctx, vdc_required=False) -> None:
         client.rehydrate_from_token(
             profiles.get('token'), profiles.get('is_jwt_token'))
 
-        ctx.obj = {}
+        ctx.obj = dict()
         ctx.obj['client'] = client
 
     _override_client(ctx)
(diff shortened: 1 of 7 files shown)


Anirudh9794:
___
*[container_service_extension/client/utils.py, line 100 at r1](https://reviewable.io/reviews/vmware/container-service-extension/690#-MEYM4Zx-feV-AUONnrA:-MEYbnaREJaMooC-wKhb:b-koxinv) ([raw file](https://github.com/vmware/container-service-extension/blob/60661b2e241dede7629d2afb4418cfdf1918fa3c/container_service_extension/client/utils.py#L100)):*

This logic is confusing. We are in a way doing below.

if is_cse_server_running:
       profiles.set(CSE_SERVER_RUNNING, True)

By the way, where are we re-setting CSE_SERVER_RUNNING flag as soon as login?
If we do not do this, once CSE_SERVER_RUNNING is set to false, it will remain false forever, right?


</blockquote></details>

In the beginning, CSE_SERVER_RUNNING will not be present. As a result, is_cse_server_running will be defaulted to True.

Profiles.yaml is flushed out when login is done by the user and new values are set. We don't have to take care of this.

container_service_extension/client/utils.py
```diff
@@ -69,14 +87,32 @@ def _override_client(ctx) -> None:
...
+            system = syst.System(ctx.obj['client'])
+            sys_info = system.get_info()
+            cse_server_api_version = sys_info.get(CSE_SERVER_API_VERSION)
+            profiles.set(CSE_SERVER_API_VERSION, cse_server_api_version)
+            profiles.set(CSE_SERVER_RUNNING, True)
+            profiles.save()
+        except Exception:
+            # If request to CSE server times out
+            profiles.set(CSE_SERVER_RUNNING, False)
+            # enable CLI for only TKG operations
```
sahithi, line 111: nit: comment # enable -> # restrict

container_service_extension/client/utils.py
```diff
@@ -7,9 +7,27 @@
 from vcd_cli.profiles import Profiles
 
 from container_service_extension.client import system as syst
+from container_service_extension.client.constants import CSE_SERVER_RUNNING
 import container_service_extension.def_.utils as def_utils
 from container_service_extension.shared_constants import CSE_SERVER_API_VERSION
 
+_RESTRICT_CLI_TO_TKG_OPERATIONS = False
```
sahithi, line 14: If true, can we show the only _cluster apply, info, delete and list commands_? It should be straight-forward, please check. If not, you can skip it.

(sorry for the late comment, it just occurred to me).

Anirudh9794:
___
*[container_service_extension/client/cse.py, line 246 at r5](https://reviewable.io/reviews/vmware/container-service-extension/690#-MExU5ka6JF0q0VUogsC:-MEyLgXx8h2Ed6lKCL6X:b-896fix) ([raw file](https://github.com/vmware/container-service-extension/blob/a7616b676dbd0201bd1b81571b38d1bb62931ef0/container_service_extension/client/cse.py#L246)):*

As discussed, please override str method to print this default message if no message is passed.
</blockquote></details>

Done.

PR #698, opened 2020-08-13
[VCDA 1653] Avoid using prefix while creating/ updating clusters

Anirudh9794:

* Avoid adding the cse---- prefix before searching for the sizing policy during adding nodes

* Testing done:
  * create a cluster with no sizing policy
  * create a cluster with with sizing policy

@sahithi @rocknes @sakthisunda

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/698)

### DIFF
--- a/container_service_extension/compute_policy_manager.py
+++ b/container_service_extension/compute_policy_manager.py
@@ -625,7 +625,10 @@ class ComputePolicyManager:
         :return: URN of the format - urn:vcloud:vdc:[VDC_ID]
         :rtype: str
         """
-        return f"{cloudapi_constants.CLOUDAPI_URN_PREFIX}:vdc:{vdc_id}"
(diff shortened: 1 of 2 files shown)


Anirudh9794:
___
*[container_service_extension/def_/cluster_service.py, line 1765 at r1](https://reviewable.io/reviews/vmware/container-service-extension/698#-MEgUjM09cnivY59Xb8T:-MF1vyaK0_XTO3U84rRt:b-iw0g2u) ([raw file](https://github.com/vmware/container-service-extension/blob/e3e31273b0ca52c8461dc7ffe440c7cbf0a30848/container_service_extension/def_/cluster_service.py#L1765)):*

If there are multiple sizing policy with the same name shouldn't multiple record exception be thrown from here?

Also can policies be None? In that case the for loop will raise exception.
</blockquote></details>

Addressed the above concern by converting the generator to a list

PR #708, opened 2020-08-21
[VCDA-1772, VCDA-1734, fix snake case output] Raise exception if no rights are present, print only task href on console

Anirudh9794:

Fixes for the following included:
* Exception with message "User cannot access native or TKG clusters. Please contact administrator" raised when user doesnt have native or tkg rights (helps in cluster list and info)
* Console output for cluster write operations (upgrade, apply and delete) will print a task_href on the console instead of the whole cluster entity
* TKG cluster info output used to return snake case. The output from defined entities should be shown as is (which is in camel case)
   - There is no need for additional data parameter which was introduced to send in tkg defined entity details. Have removed it from the tkg client

Testing done:
TKG cluster:
* list, cluster info , delete and apply commands to see if the response is proper
Native cluster ( TODO )
* list, cluster info, delete, and apply commands and checked if the response is proper

* created user with no rights and saw if the appropriate error message is shown on the screen (TODO)

Screenshots:
list and info TKG cluster
Screen Shot 2020-08-21 at 1 14 49 PM

delete TKG cluster
Screen Shot 2020-08-21 at 1 17 20 PM

@sahithi @sakthisunda @rocknes

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/708)

### DIFF
--- a/container_service_extension/client/def_entity_cluster_api.py
+++ b/container_service_extension/client/def_entity_cluster_api.py
@@ -4,7 +4,7 @@
 from dataclasses import asdict
 import os
 
-from requests.exceptions import HTTPError
+import requests
 import yaml
 
 import container_service_extension.client.constants as cli_constants
@@ -73,7 +73,7 @@ class DefEntityClusterApi:
         try:
             clusters += self._tkgCluster.list_tkg_clusters(vdc=vdc, org=org)
         except tkg_rest.ApiException as e:
-            if e.status != 403:
+            if e.status != requests.codes.FORBIDDEN:
                 raise
             has_tkg_rights = False
         try:
@@ -90,8 +90,8 @@ class DefEntityClusterApi:
                     cli_constants.CLIOutputKey.OWNER.value: def_entity.owner.name  # noqa: E501
                 }
                 clusters.append(cluster)
-        except HTTPError as e:
-            if e.response.status_code != 403:
+        except requests.exceptions.HTTPError as e:
+            if e.response.status_code != requests.codes.FORBIDDEN:
                 raise
             has_native_rights = False
         if not (has_tkg_rights and has_native_rights):
@@ -141,7 +141,7 @@ class DefEntityClusterApi:
                 vdc=vdc,
                 org=org)
         except tkg_rest.ApiException as e:
-            if e.status != 403:
+            if e.status != requests.codes.FORBIDDEN:
                 raise
             has_tkg_rights = False
         if not (has_native_rights or has_tkg_rights):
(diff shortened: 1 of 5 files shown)


Anirudh9794:
___
*[container_service_extension/client/def_entity_cluster_api.py, line 149 at r1](https://reviewable.io/reviews/vmware/container-service-extension/708#-MFWlfuj9eWjQO8t9DTh:-MFWnv0h17b_KoRwFUSB:b-wzk8j7) ([raw file](https://github.com/vmware/container-service-extension/blob/eed51e9f97b8e61f565b7d005c33368944726a1b/container_service_extension/client/def_entity_cluster_api.py#L149)):*

Rpecifying the k8s runtime won't help us to narrow down to a single cluster if len(tkg_def_entities) > 1, right? Can there be multiple tkg clusters with the same name? If yes, how do we filter them down to just 1 cluster?
</blockquote></details>

I discussed this with Sahithi, Name conflict can occur among 2 TKG clusters in different organizations, TKG and native clusters in the same organization and also TKG and native clusters in different organization. Based on this, we have decided the following approach:
1. If there is name conflict, first check if 'org' parameter is set. If not set, Ask the user to specify the org.
2. If org is specified, ask the user to specify k8-runtime parameter as well.

Anirudh9794:
___
*[container_service_extension/client/def_entity_cluster_api.py, line 176 at r1](https://reviewable.io/reviews/vmware/container-service-extension/708#-MFWm61l7WyEUJshlGwl:-MFho2cnEVeIuMPPu_Y3:b-ie4wil) ([raw file](https://github.com/vmware/container-service-extension/blob/eed51e9f97b8e61f565b7d005c33368944726a1b/container_service_extension/client/def_entity_cluster_api.py#L176)):*

This if block of code is being repeated in few places. Can't this be consolidated in a single method?

_get_tkg_native_clusters_by_name is never being used in a way where we expect more than 1 cluster result. We can put the if check in this method and return the dict + a boolean to specify whether the cluster is of type native or tkg?

Other suggestions are welcome too.
</blockquote></details>

Changed it as per the above suggestion

Anirudh9794:
___
*[container_service_extension/client/def_entity_cluster_api.py, line 155 at r3](https://reviewable.io/reviews/vmware/container-service-extension/708#-MFmXZ0g-PaX9BFZsrWr:-MFmXZ0g-PaX9BFZsrWs:bloqgnc) ([raw file](https://github.com/vmware/container-service-extension/blob/bd77dab528cfc552e9460bc9c89718f754d4de8c/container_service_extension/client/def_entity_cluster_api.py#L155)):*
> ```Python
>             # filter
>             if not org:
>                 # handles the case where there more than 1 TKG cluster with
> ```

case where there is TKG and native cluser in the same org

PR #785, opened 2020-10-13
[VCDA-1877] Consumer watchdog

ltimothy7:
Signed-off-by: Timothy Le <ltimothy@vmware.com>

- (Required) Short description of changes in the PR topic line
This PR includes adding a consumer watchdog, editing the shutdown logic for MQTT and AMQP consumers, and updated the server busy logic.

-Consumer watchdog recreates consumer if the consumer thread is not alive
-MQTT and AMQP consumers shut down appropriately
-MQTT and AMQP: receive server new busy message when all threads are busy

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/785)


ltimothy7 (commented):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/785)* status: 0 of 6 files reviewed, 1 unresolved discussion

___
*[container_service_extension/consumer/amqp_consumer.py, line 245 at r2](https://reviewable.io/reviews/vmware/container-service-extension/785#-MJXqqR66eE7fpB1Ofmq:-MJXqqR66eE7fpB1Ofmr:b-y7aqw) ([raw file](https://github.com/vmware/container-service-extension/blob/fc4f2ebe00529b966cee332620632a6a89b5b010/container_service_extension/consumer/amqp_consumer.py#L245)):*
> ```Python
>         self.stop_consuming()
>         if self._connection:
>             self._connection.ioloop.stop()
> ```

I think that this logic previously followed the following example: https://pika.readthedocs.io/en/0.13.1/examples/asynchronous_consumer_example.html 

This example called ioloop.start() when the consumer was closing because the consumer was running in the main thread; the thread's ioloop.start() was interrupted during a KeyboardInterrupt, so ioloop.start() during close to restart the connection and it would close on its own. 
In our use case, the consumer is a separately spawned thread, so we do not call ioloop.start() during close sine this ioloop was not interrupted, but we close the connection.

sakthisunda (commented):
> Reviewed 6 of 10 files at r2.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/785)* status: all files reviewed, 2 unresolved discussions (waiting on @ltimothy7)

___
*[container_service_extension/service.py, line 77 at r2](https://reviewable.io/reviews/vmware/container-service-extension/785#-MJbr2uECh9oWetyYnp6:-MJbr2uECh9oWetyYnp7:bp9bp7d) ([raw file](https://github.com/vmware/container-service-extension/blob/fc4f2ebe00529b966cee332620632a6a89b5b010/container_service_extension/service.py#L77)):*
> ```Python
> 
> 
> def watchdog_thread_run(service_obj, num_processors):
> ```

Curious to know: what is the lifecycle of watchdog thread. Do we need a shut down explicit?

sakthisunda (approved):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/785)* status: all files reviewed, 2 unresolved discussions (waiting on @ltimothy7)

sakthisunda (commented):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/785)* status: all files reviewed, 3 unresolved discussions (waiting on @ltimothy7)

___
*[container_service_extension/service.py, line 78 at r2](https://reviewable.io/reviews/vmware/container-service-extension/785#-MJcagPf5etyR-Cj949T:-MJcagPf5etyR-Cj949U:b-eme5xj) ([raw file](https://github.com/vmware/container-service-extension/blob/fc4f2ebe00529b966cee332620632a6a89b5b010/container_service_extension/service.py#L78)):*
> ```Python
> 
> def watchdog_thread_run(service_obj, num_processors):
>     logger.SERVER_LOGGER.info("Starting watchdog thread")
> ```

logging which watchdog thread id(id/name) that watches which consumer id/name would be useful.

ltimothy7 (commented):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/785)* status: all files reviewed, 3 unresolved discussions (waiting on @sakthisunda)

___
*[container_service_extension/service.py, line 77 at r2](https://reviewable.io/reviews/vmware/container-service-extension/785#-MJbr2uECh9oWetyYnp6:-MJgl57l1qEEwiPoS4j5:b-koq5yp) ([raw file](https://github.com/vmware/container-service-extension/blob/fc4f2ebe00529b966cee332620632a6a89b5b010/container_service_extension/service.py#L77)):*

Curious to know: what is the lifecycle of watchdog thread. Do we need a shut down explicit? 
</blockquote></details>

The watchdog thread is a daemon thread, so it will exit with the program
I also added a check for the loop to break if the service status is not "running" so that the watchdog thread can also exit on its own.
___
*[container_service_extension/service.py, line 78 at r2](https://reviewable.io/reviews/vmware/container-service-extension/785#-MJcagPf5etyR-Cj949T:-MJgrHq3205BYjwSG_xy:b-6eo87z) ([raw file](https://github.com/vmware/container-service-extension/blob/fc4f2ebe00529b966cee332620632a6a89b5b010/container_service_extension/service.py#L78)):*

logging which watchdog thread id(id/name) that watches which consumer id/name would be useful. 
</blockquote></details>

I have added this logic to the run method similar to the message consumer logging.

rocknes (changes requested):
> Reviewed 3 of 10 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/785)* status: all files reviewed, 7 unresolved discussions (waiting on @ltimothy7 and @sakthisunda)

___
*[container_service_extension/server_constants.py, line 71 at r3](https://reviewable.io/reviews/vmware/container-service-extension/785#-MK19DIc3DAbpzHHu6W2:-MK19DIc3DAbpzHHu6W3:b-o1jo0h) ([raw file](https://github.com/vmware/container-service-extension/blob/d0776b57c14a5a96a7ed49ca15287cf1ffdad499/container_service_extension/server_constants.py#L71)):*
> ```Python
> 
> # CSE Server Busy strings
> CSE_SERVER_BUSY_KEY = 'CSE Server Busy'
> ```

Please move this to shared_constants.py since both client and server are accessing this constant.
___
*[container_service_extension/service.py, line 80 at r3](https://reviewable.io/reviews/vmware/container-service-extension/785#-MK17D2D8Uwct6-zYRrR:-MK17D2ED5jUhPhFRAl4:bot4q5p) ([raw file](https://github.com/vmware/container-service-extension/blob/d0776b57c14a5a96a7ed49ca15287cf1ffdad499/container_service_extension/service.py#L80)):*
> ```Python
>     logger.SERVER_LOGGER.info("Starting watchdog thread")
>     while True:
>         if service_obj.get_status() != ServerState.RUNNING.value:
> ```

This means that if the server is disabled via `vcd cse system disable` the watchdog thread will exit.

I think the logic should be to exit only if the service is in "STOPPED" state.
If the service is in "RUNNING" state and the consumer dies , then restart it.
If the service is in "DISABLED/STOPPING" state then don't restart the consumer but keep the watchdog thread alive.
___
*[container_service_extension/client/utils.py, line 194 at r3](https://reviewable.io/reviews/vmware/container-service-extension/785#-MK18epx5yWYP5pYeAPp:-MK18epx5yWYP5pYeAPq:b-qvdkcv) ([raw file](https://github.com/vmware/container-service-extension/blob/d0776b57c14a5a96a7ed49ca15287cf1ffdad499/container_service_extension/client/utils.py#L194)):*
> ```Python
>         return filtered_result
>     elif isinstance(result, dict):
>         # Check if result is CSE Server busy message
> ```

Please add comment to clarify that since the server returned busy message, there is no result to filter and hence this function returns the original error message as is.
___
*[container_service_extension/consumer/mqtt_consumer.py, line 163 at r3](https://reviewable.io/reviews/vmware/container-service-extension/785#-MK1Aqh43iT9DcQgXrmO:-MK1Aqh43iT9DcQgXrmP:b-uk8ago) ([raw file](https://github.com/vmware/container-service-extension/blob/d0776b57c14a5a96a7ed49ca15287cf1ffdad499/container_service_extension/consumer/mqtt_consumer.py#L163)):*
> ```Python
>         self._is_closing = True
>         self._ctpe.shutdown(wait=True)  # Let jobs finish before disconnecting
>         if self._mqtt_client:
> ```

Do we have to call stop_consuming too, like we do for amqp?

rocknes (commented):
> Reviewed 1 of 10 files at r2.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/785)* status: all files reviewed, 7 unresolved discussions (waiting on @ltimothy7 and @sakthisunda)

ltimothy7 (commented):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/785)* status: all files reviewed, 7 unresolved discussions (waiting on @rocknes and @sakthisunda)

___
*[container_service_extension/server_constants.py, line 71 at r3](https://reviewable.io/reviews/vmware/container-service-extension/785#-MK19DIc3DAbpzHHu6W2:-MK6DTgO5Q7-BgGp6d7B:b-896fix) ([raw file](https://github.com/vmware/container-service-extension/blob/d0776b57c14a5a96a7ed49ca15287cf1ffdad499/container_service_extension/server_constants.py#L71)):*

Please move this to shared_constants.py since both client and server are accessing this constant.
</blockquote></details>

Done.
___
*[container_service_extension/service.py, line 80 at r3](https://reviewable.io/reviews/vmware/container-service-extension/785#-MK17D2D8Uwct6-zYRrR:-MK6E7E31dm338IMqhVM:bxazkjk) ([raw file](https://github.com/vmware/container-service-extension/blob/d0776b57c14a5a96a7ed49ca15287cf1ffdad499/container_service_extension/service.py#L80)):*

This means that if the server is disabled via `vcd cse system disable` the watchdog thread will exit.

I think the logic should be to exit only if the service is in "STOPPED" state.
If the service is in "RUNNING" state and the consumer dies , then restart it.
If the service is in "DISABLED/STOPPING" state then don't restart the consumer but keep the watchdog thread alive.
</blockquote></details>

Done. Thank you for this catch
___
*[container_service_extension/client/utils.py, line 194 at r3](https://reviewable.io/reviews/vmware/container-service-extension/785#-MK18epx5yWYP5pYeAPp:-MK6EuJa7cgOMOfUla-K:b-896fix) ([raw file](https://github.com/vmware/container-service-extension/blob/d0776b57c14a5a96a7ed49ca15287cf1ffdad499/container_service_extension/client/utils.py#L194)):*

Please add comment to clarify that since the server returned busy message, there is no result to filter and hence this function returns the original error message as is.
</blockquote></details>

Done.
___
*[container_service_extension/consumer/mqtt_consumer.py, line 163 at r3](https://reviewable.io/reviews/vmware/container-service-extension/785#-MK1Aqh43iT9DcQgXrmO:-MK6F44KCOjXKeWG_Lch:b-ougnil) ([raw file](https://github.com/vmware/container-service-extension/blob/d0776b57c14a5a96a7ed49ca15287cf1ffdad499/container_service_extension/consumer/mqtt_consumer.py#L163)):*

Do we have to call stop_consuming too, like we do for amqp?
</blockquote></details>

We don't have to. AMQP's stop_consuming is specific to the AMQP message protocol in which both the channel and connection must be canceled/closed. For MQTT, we just need to disconnect from the broker.

rocknes (approved):
:lgtm:

> Reviewed 5 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/785#-:-MK6NyrG8T0fRFWNemJK:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @sakthisunda)

PR #851, opened 2020-12-11
[VCDA-1909] Add logic in AMQP consumer to reject redelivered messages

rocknes:
CSE today is incapable to handling redelivered messages. This can lead to situations where CSE reprocesses heavy incoming requests multiple times. To make CSE server resilient to redelivered message, in this PR we have added logic to detect duplicate requests and reject them.

Testing done:
As part of testing added an exception in the normal workflow to crash the consumer to guarantee a message redelivery. Checked the logs and made sure that on redelivery of the message it is rejected, as well as removed from the AMQP queue.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/851)

### DIFF
--- a/container_service_extension/consumer/amqp_consumer.py
+++ b/container_service_extension/consumer/amqp_consumer.py
@@ -18,7 +18,8 @@ import container_service_extension.consumer.utils as utils
 from container_service_extension.logger import SERVER_LOGGER as LOGGER
 from container_service_extension.server_constants import EXCHANGE_TYPE
 
-REQUESTS_BEING_PROCESSSED = LRU(constants.MAX_PROCESSING_REQUEST_CACHE_SIZE)
+REQUESTS_BEING_PROCESSED = LRU(constants.MAX_PROCESSING_REQUEST_CACHE_SIZE)
+LRU_LOCK = Lock()
 
 
 class AMQPConsumer(object):
@@ -189,9 +190,13 @@ class AMQPConsumer(object):
             self.send_response(reply_msg, properties)
             LOGGER.debug(f"Sucessfully sent reply: {reply_msg} to AMQP.")
 
-        global REQUESTS_BEING_PROCESSSED
-        if req_id in REQUESTS_BEING_PROCESSSED:
-            del REQUESTS_BEING_PROCESSSED[req_id]
+        global REQUESTS_BEING_PROCESSED, LRU_LOCK
+        LRU_LOCK.acquire()
+        try:
+            if req_id in REQUESTS_BEING_PROCESSED:
+                del REQUESTS_BEING_PROCESSED[req_id]
+        finally:
+            LRU_LOCK.release()
 
     def send_response(self, reply_msg, properties):
         reply_properties = pika.BasicProperties(
@@ -228,17 +233,25 @@ class AMQPConsumer(object):
 
         req_id = utils.get_request_id(
             request_msg=body, fsencoding=self.fsencoding)
-        global REQUESTS_BEING_PROCESSSED
-        if req_id in REQUESTS_BEING_PROCESSSED:
-            self.reject_message(basic_deliver.delivery_tag)
-            del REQUESTS_BEING_PROCESSSED[req_id]
-            return
+        global REQUESTS_BEING_PROCESSED, LRU_LOCK
+        LRU_LOCK.acquire()
+        try:
+            if req_id in REQUESTS_BEING_PROCESSED:
+                self.reject_message(basic_deliver.delivery_tag)
+                del REQUESTS_BEING_PROCESSED[req_id]
+                return
+        finally:
+            LRU_LOCK.release()
 
         self.acknowledge_message(basic_deliver.delivery_tag)
         if self._ctpe.max_threads_busy():
             self.send_too_many_requests_response(properties, body)
         else:
-            REQUESTS_BEING_PROCESSSED[req_id] = time.time()
+            LRU_LOCK.acquire()
+            try:
+                REQUESTS_BEING_PROCESSED[req_id] = time.time()
+            finally:
+                LRU_LOCK.release()
             self._ctpe.submit(lambda: self.process_amqp_message(
                 properties, body, basic_deliver))
 
--- a/container_service_extension/consumer/amqp_consumer.py
+++ b/container_service_extension/consumer/amqp_consumer.py
@@ -5,7 +5,6 @@
 import json
 import sys
 from threading import Lock
-import time
 
 from lru import LRU
 import pika
@@ -249,7 +248,7 @@ class AMQPConsumer(object):
         else:
             LRU_LOCK.acquire()
             try:
-                REQUESTS_BEING_PROCESSED[req_id] = time.time()
+                REQUESTS_BEING_PROCESSED[req_id] = True
             finally:
                 LRU_LOCK.release()
             self._ctpe.submit(lambda: self.process_amqp_message(
--- a/container_service_extension/consumer/amqp_consumer.py
+++ b/container_service_extension/consumer/amqp_consumer.py
@@ -189,7 +189,12 @@ class AMQPConsumer(object):
                 reply_body_str=reply_body_str)
 
             self.send_response(reply_msg, properties)
-            LOGGER.debug(f"AMQP reply: {reply_msg}")
+            LOGGER.debug(f"Sucessfully sent reply: {reply_msg} to AMQP.")
+            if req_id in REQUESTS_BEING_PROCESSSED:
+                del REQUESTS_BEING_PROCESSSED[req_id]
+        else:
+            LOGGER.debug(f"Unable to reply to message {delivery_tag}")
+            self.reject_message(basic_deliver.delivery_tag)
 
     def send_response(self, reply_msg, properties):
         reply_properties = pika.BasicProperties(
@@ -228,7 +233,7 @@ class AMQPConsumer(object):
             request_msg=body, fsencoding=self.fsencoding)
         global REQUESTS_BEING_PROCESSSED
         if req_id not in REQUESTS_BEING_PROCESSSED:
-            REQUESTS_BEING_PROCESSSED[req_id] = time.time()
+            REQUESTS_BEING_PROCESSSED.set(req_id, time.time())
             self.acknowledge_message(basic_deliver.delivery_tag)
         else:
             self.reject_message(basic_deliver.delivery_tag)
--- a/container_service_extension/consumer/amqp_consumer.py
+++ b/container_service_extension/consumer/amqp_consumer.py
@@ -5,7 +5,9 @@
 import json
 import sys
 from threading import Lock
+import time
 
+from lru import LRU
 import pika
 import requests
 
@@ -15,6 +17,10 @@ from container_service_extension.consumer.consumer_thread_pool_executor \
 import container_service_extension.consumer.utils as utils
 from container_service_extension.logger import SERVER_LOGGER as LOGGER
 from container_service_extension.server_constants import EXCHANGE_TYPE
+
+
+MAX_PROCESSED_REQUEST_CACHE_SIZE = 1000
+REQUESTS_BEING_PROCESSSED = LRU(MAX_PROCESSED_REQUEST_CACHE_SIZE)
 
 
 class AMQPConsumer(object):
@@ -218,7 +224,16 @@ class AMQPConsumer(object):
         if channel.is_closed or channel.is_closing:
             return
 
-        self.acknowledge_message(basic_deliver.delivery_tag)
+        req_id = utils.get_request_id(
+            request_msg=body, fsencoding=self.fsencoding)
+        global REQUESTS_BEING_PROCESSSED
+        if req_id not in REQUESTS_BEING_PROCESSSED:
+            REQUESTS_BEING_PROCESSSED[req_id] = time.time()
+            self.acknowledge_message(basic_deliver.delivery_tag)
+        else:
+            self.reject_message(basic_deliver.delivery_tag)
+            return
+
         if self._ctpe.max_threads_busy():
             self.send_too_many_requests_response(properties, body)
         else:
@@ -229,6 +244,11 @@ class AMQPConsumer(object):
     def acknowledge_message(self, delivery_tag):
         LOGGER.debug(f"Acknowledging message ({delivery_tag})")
         self._channel.basic_ack(delivery_tag=delivery_tag)
+
+    def reject_message(self, delivery_tag):
+        LOGGER.debug(f"Rejecting message {delivery_tag}")
+        self._channel.basic_nack(
+            delivery_tag=delivery_tag, requeue=False)
 
     def stop_consuming(self):
         if self._channel:
--- a/container_service_extension/consumer/amqp_consumer.py
+++ b/container_service_extension/consumer/amqp_consumer.py
@@ -18,9 +18,7 @@ import container_service_extension.consumer.utils as utils
 from container_service_extension.logger import SERVER_LOGGER as LOGGER
 from container_service_extension.server_constants import EXCHANGE_TYPE
 
-
-MAX_PROCESSED_REQUEST_CACHE_SIZE = 1000
-REQUESTS_BEING_PROCESSSED = LRU(MAX_PROCESSED_REQUEST_CACHE_SIZE)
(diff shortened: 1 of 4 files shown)


ltimothy7 (commented):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/851)* status: all files reviewed, 2 unresolved discussions (waiting on @rocknes)

___
*[container_service_extension/consumer/amqp_consumer.py, line 22 at r1](https://reviewable.io/reviews/vmware/container-service-extension/851#-MOI55N6-qbjZJ7avQC5:-MOI55N6-qbjZJ7avQC6:b-vk7j6s) ([raw file](https://github.com/vmware/container-service-extension/blob/da7ea266e1552d72ef0090e075381eaaedb94f08/container_service_extension/consumer/amqp_consumer.py#L22)):*
> ```Python
> 
> 
> MAX_PROCESSED_REQUEST_CACHE_SIZE = 1000
> ```

nit: I think this should be moved to consumer/constants.py
___
*[container_service_extension/consumer/amqp_consumer.py, line 220 at r1](https://reviewable.io/reviews/vmware/container-service-extension/851#-MOI5WZ9BUVATJFbIG61:-MOI5WZ9BUVATJFbIG62:b3w3ch6) ([raw file](https://github.com/vmware/container-service-extension/blob/da7ea266e1552d72ef0090e075381eaaedb94f08/container_service_extension/consumer/amqp_consumer.py#L220)):*
> ```Python
>         global REQUESTS_BEING_PROCESSSED
>         if req_id not in REQUESTS_BEING_PROCESSSED:
>             REQUESTS_BEING_PROCESSSED[req_id] = time.time()
> ```

Should the time stored in REQUESTS_BEING_PROCESSED be used to determine whether or not to acknowledge the message?

ChintanShahVMware:
___
*[container_service_extension/consumer/amqp_consumer.py, line 22 at r1](https://reviewable.io/reviews/vmware/container-service-extension/851#-MOI8pXg9wo7euzKr_qN:-MOI8pXg9wo7euzKr_qO:byftj7g) ([raw file](https://github.com/vmware/container-service-extension/blob/da7ea266e1552d72ef0090e075381eaaedb94f08/container_service_extension/consumer/amqp_consumer.py#L22)):*
> ```Python
> 
> 
> MAX_PROCESSED_REQUEST_CACHE_SIZE = 1000
> ```

Whats the reasoning behind setting LRU cache size to 1000?

ChintanShahVMware:
___
*[container_service_extension/consumer/amqp_consumer.py, line 22 at r1](https://reviewable.io/reviews/vmware/container-service-extension/851#-MOI8pXg9wo7euzKr_qN:-MOI8zFl5FCwCklsG1KW:b-v5pbbs) ([raw file](https://github.com/vmware/container-service-extension/blob/da7ea266e1552d72ef0090e075381eaaedb94f08/container_service_extension/consumer/amqp_consumer.py#L22)):*
<details><summary><i>Previously, ChintanShahVMware wrote…</i></summary><blockquote>

Whats the reasoning behind setting LRU cache size to 1000?
</blockquote></details>

I mean, what if a customer has 100+ service threads (might be possible in case of CSE on CDS). We can possibly lose the message-ids in this case.
IMO, we need to come-up with an equation here based on:
Figure out how many requests are processed by a service thread in a single second.
Whats the duration (config) in which AMQP will enqueue message again.
What are the number of threads.

keeping it fixed to 1000 might lead to wrong results.

ChintanShahVMware:
___
*[container_service_extension/consumer/amqp_consumer.py, line 220 at r1](https://reviewable.io/reviews/vmware/container-service-extension/851#-MOI5WZ9BUVATJFbIG61:-MOI9uOQEu5SI_u6wbb3:bq4nubu) ([raw file](https://github.com/vmware/container-service-extension/blob/da7ea266e1552d72ef0090e075381eaaedb94f08/container_service_extension/consumer/amqp_consumer.py#L220)):*

Should the time stored in REQUESTS_BEING_PROCESSED be used to determine whether or not to acknowledge the message?
</blockquote></details>

+1 to this.

rocknes (commented):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/851)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @ChintanShahVMware and @ltimothy7)

___
*[container_service_extension/consumer/amqp_consumer.py, line 22 at r1](https://reviewable.io/reviews/vmware/container-service-extension/851#-MOI55N6-qbjZJ7avQC5:-MOYYg9b853XBuFvUt8R:b-ds0ka6) ([raw file](https://github.com/vmware/container-service-extension/blob/da7ea266e1552d72ef0090e075381eaaedb94f08/container_service_extension/consumer/amqp_consumer.py#L22)):*

nit: I think this should be moved to consumer/constants.py
</blockquote></details>

sure, will do that.
___
*[container_service_extension/consumer/amqp_consumer.py, line 22 at r1](https://reviewable.io/reviews/vmware/container-service-extension/851#-MOI8pXg9wo7euzKr_qN:-MOYYixv6u-s9yry-64a:b-64ckho) ([raw file](https://github.com/vmware/container-service-extension/blob/da7ea266e1552d72ef0090e075381eaaedb94f08/container_service_extension/consumer/amqp_consumer.py#L22)):*
<details><summary><i>Previously, ChintanShahVMware wrote…</i></summary><blockquote>

I mean, what if a customer has 100+ service threads (might be possible in case of CSE on CDS). We can possibly lose the message-ids in this case.
IMO, we need to come-up with an equation here based on:
Figure out how many requests are processed by a service thread in a single second.
Whats the duration (config) in which AMQP will enqueue message again.
What are the number of threads.

keeping it fixed to 1000 might lead to wrong results.
</blockquote></details>

The things that you have mentioned are configuration parameters not available to CSE for most cases. This particular fix won't affect everyday message processing. Only the most pathological cases will hit this fix.

The number 1000 is pretty large to be honest. AMQP message  re-delivery on failure generally happens every minute to 2 minute. Typically CSE runs with just 15 threads and we have no published recommended numbers, so I find it hard to believe that providers would run it with 100 threads. Even in case of telefonica (who has around 1000+ ovdc) were running with just 10 consumer threads.

And from the scale and perf numbers that SUPI team found out, turn around time for CSE is only about 6 requests /min (that includes the fast GET calls). So we have a headroom of about 100 requests per thread (or about 17 minutes) before we hit any significant issue with the cache.

But I see your point, let me look into the library and see if they have a force eviction policy. In that case completed requests can be evicted, leaving only the requests that are being processed. When a request is successfully processed and response is acknowledged by AMQP, the message will never be re-delivered.
___
*[container_service_extension/consumer/amqp_consumer.py, line 220 at r1](https://reviewable.io/reviews/vmware/container-service-extension/851#-MOI5WZ9BUVATJFbIG61:-MOY_eeGD4yIHm0qBMxK:bpnf04q) ([raw file](https://github.com/vmware/container-service-extension/blob/da7ea266e1552d72ef0090e075381eaaedb94f08/container_service_extension/consumer/amqp_consumer.py#L220)):*
<details><summary><i>Previously, ChintanShahVMware wrote…</i></summary><blockquote>

+1 to this.
</blockquote></details>

Wanted to use it for some time based rejection, unfortunately the AMQP message payload doesn't have a field for incoming request time :|
At this point this dict is acting more like a set. I can change the timestamp to just a plain boolean.

Anirudh9794:
___
*[container_service_extension/consumer/amqp_consumer.py, line 192 at r3](https://reviewable.io/reviews/vmware/container-service-extension/851#-MOiGNat4Ep3Lj2-23HI:-MOiGNat4Ep3Lj2-23HJ:bxqpppn) ([raw file](https://github.com/vmware/container-service-extension/blob/187615c341b68e487bdf27ea43d4dca7850a7c5a/container_service_extension/consumer/amqp_consumer.py#L192)):*
> ```Python
>             LOGGER.debug(f"Sucessfully sent reply: {reply_msg} to AMQP.")
> 
>         global REQUESTS_BEING_PROCESSSED
> ```

Since this variable is shared between threads, I think it should be guarded by a mutex. I don't think it will cause a big problem as one or the other duplicated request will be rejected. But I feel this variable needs a lock since it can possibly crash the CSE server

ChintanShahVMware:
Other than Aniruddha's most recent comment regarding a lock on shared LRU cache, everything else looks good to me.

ChintanShahVMware:
___
*[container_service_extension/consumer/amqp_consumer.py, line 22 at r1](https://reviewable.io/reviews/vmware/container-service-extension/851#-MOI8pXg9wo7euzKr_qN:-MOidWzoCI6jtlcrnuuF:bq009j9) ([raw file](https://github.com/vmware/container-service-extension/blob/da7ea266e1552d72ef0090e075381eaaedb94f08/container_service_extension/consumer/amqp_consumer.py#L22)):*

The things that you have mentioned are configuration parameters not available to CSE for most cases. This particular fix won't affect everyday message processing. Only the most pathological cases will hit this fix.

The number 1000 is pretty large to be honest. AMQP message  re-delivery on failure generally happens every minute to 2 minute. Typically CSE runs with just 15 threads and we have no published recommended numbers, so I find it hard to believe that providers would run it with 100 threads. Even in case of telefonica (who has around 1000+ ovdc) were running with just 10 consumer threads.

And from the scale and perf numbers that SUPI team found out, turn around time for CSE is only about 6 requests /min (that includes the fast GET calls). So we have a headroom of about 100 requests per thread (or about 17 minutes) before we hit any significant issue with the cache.

But I see your point, let me look into the library and see if they have a force eviction policy. In that case completed requests can be evicted, leaving only the requests that are being processed. When a request is successfully processed and response is acknowledged by AMQP, the message will never be re-delivered.
</blockquote></details>

Got it.
Agreed, for this type of workload, i think we are fine with hard-coding cache size.
Although lets just document it as a comment before cache-size-declaration, else it will look like a magic-number.

arunmk (commented):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/851)* status: 1 of 4 files reviewed, 5 unresolved discussions (waiting on @Anirudh9794, @ltimothy7, and @rocknes)

___
*[container_service_extension/consumer/amqp_consumer.py, line 21 at r3](https://reviewable.io/reviews/vmware/container-service-extension/851#-MOirqrf-7BgnRLS75IV:-MOirqrf-7BgnRLS75IW:bebt6kl) ([raw file](https://github.com/vmware/container-service-extension/blob/187615c341b68e487bdf27ea43d4dca7850a7c5a/container_service_extension/consumer/amqp_consumer.py#L21)):*
> ```Python
> from container_service_extension.server_constants import EXCHANGE_TYPE
> 
> REQUESTS_BEING_PROCESSSED = LRU(constants.MAX_PROCESSING_REQUEST_CACHE_SIZE)
> ```

nit: typo PROCESSSED
___
*[container_service_extension/consumer/amqp_consumer.py, line 192 at r3](https://reviewable.io/reviews/vmware/container-service-extension/851#-MOiGNat4Ep3Lj2-23HI:-MOiqE5A9mXqOEmFeb_d:bng932n) ([raw file](https://github.com/vmware/container-service-extension/blob/187615c341b68e487bdf27ea43d4dca7850a7c5a/container_service_extension/consumer/amqp_consumer.py#L192)):*

Since this variable is shared between threads, I think it should be guarded by a mutex. I don't think it will cause a big problem as one or the other duplicated request will be rejected. But I feel this variable needs a lock since it can possibly crash the CSE server
</blockquote></details>

+1 on this. Ignored during the earlier discussion since this class doesn't really handle exceptions in general. (Since this is a pure in-memory operation we would likely be protected by the gil, but it is cheap code. And it is easier to miss this when exception handling is added later on.)

arunmk (approved):
> Reviewed 1 of 3 files at r1, 2 of 3 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/851)* status: all files reviewed, 5 unresolved discussions (waiting on @Anirudh9794, @ltimothy7, and @rocknes)

rocknes (commented):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/851)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @Anirudh9794, @arunmk, @ltimothy7, and @rocknes)

___
*[container_service_extension/consumer/amqp_consumer.py, line 21 at r3](https://reviewable.io/reviews/vmware/container-service-extension/851#-MOirqrf-7BgnRLS75IV:-MOrv7yj755FXszdRkdD:b-gki9fo) ([raw file](https://github.com/vmware/container-service-extension/blob/187615c341b68e487bdf27ea43d4dca7850a7c5a/container_service_extension/consumer/amqp_consumer.py#L21)):*
<details><summary><i>Previously, arunmk (Arun M. Krishnakumar) wrote…</i></summary><blockquote>

nit: typo PROCESSSED
</blockquote></details>

thanks, fixed.
___
*[container_service_extension/consumer/amqp_consumer.py, line 192 at r3](https://reviewable.io/reviews/vmware/container-service-extension/851#-MOiGNat4Ep3Lj2-23HI:-MOrv9Sy8ni0M4Wapyz0:bm3iztu) ([raw file](https://github.com/vmware/container-service-extension/blob/187615c341b68e487bdf27ea43d4dca7850a7c5a/container_service_extension/consumer/amqp_consumer.py#L192)):*
<details><summary><i>Previously, arunmk (Arun M. Krishnakumar) wrote…</i></summary><blockquote>

+1 on this. Ignored during the earlier discussion since this class doesn't really handle exceptions in general. (Since this is a pure in-memory operation we would likely be protected by the gil, but it is cheap code. And it is easier to miss this when exception handling is added later on.)
</blockquote></details>

Added mutex lock to guard all access(es) to the LRU dict

ltimothy7 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/851)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @Anirudh9794, @arunmk, @ChintanShahVMware, and @rocknes)

___
*[container_service_extension/consumer/amqp_consumer.py, line 220 at r1](https://reviewable.io/reviews/vmware/container-service-extension/851#-MOI5WZ9BUVATJFbIG61:-MOrwvepDzS40mc4XpWS:bo5j9w5) ([raw file](https://github.com/vmware/container-service-extension/blob/da7ea266e1552d72ef0090e075381eaaedb94f08/container_service_extension/consumer/amqp_consumer.py#L220)):*

Wanted to use it for some time based rejection, unfortunately the AMQP message payload doesn't have a field for incoming request time :|
At this point this dict is acting more like a set. I can change the timestamp to just a plain boolean.
</blockquote></details>

I think changing this to have boolean values would be better with a comment saying that we are using this dict as a set. Then, the time library is not needed,

rocknes (commented):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/851)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @Anirudh9794, @arunmk, @ltimothy7, and @rocknes)

___
*[container_service_extension/consumer/amqp_consumer.py, line 220 at r1](https://reviewable.io/reviews/vmware/container-service-extension/851#-MOI5WZ9BUVATJFbIG61:-MO_Jvu39tei3nWOQK8y:b-896fix) ([raw file](https://github.com/vmware/container-service-extension/blob/da7ea266e1552d72ef0090e075381eaaedb94f08/container_service_extension/consumer/amqp_consumer.py#L220)):*

I think changing this to have boolean values would be better with a comment saying that we are using this dict as a set. Then, the time library is not needed,
</blockquote></details>

Done.

Anirudh9794 (approved):
:lgtm: 

> Reviewed 2 of 3 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/851#-:-MOs11Pw-TqHZEfImWKJ:bg7igkv)* status: all files reviewed, 4 unresolved discussions (waiting on @Anirudh9794, @arunmk, @ltimothy7, and @rocknes)

ChintanShahVMware:
:lgtm:

ltimothy7 (commented):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/851#-:-MOs4NSb633Bzi__InGO:bg7igkv)* status: all files reviewed, 4 unresolved discussions (waiting on @Anirudh9794, @arunmk, @ltimothy7, and @rocknes)

PR #877, opened 2021-01-20
[VCDA-2002] Avoid resolving defined entities after every update

Anirudh9794:

Testing done:
* `vcd cse cluster apply input_config.yaml` (create, resize up and resize down)
* `vcd cse cluster upgradel`

@sahithi @sakthisunda

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/877)

### DIFF
--- a/container_service_extension/def_/cluster_service.py
+++ b/container_service_extension/def_/cluster_service.py
@@ -766,6 +766,7 @@ class ClusterService(abstract_broker.AbstractBroker):
         except (e.ControlPlaneNodeCreationError, e.WorkerNodeCreationError,
                 e.NFSNodeCreationError, e.ClusterJoiningError,
                 e.ClusterInitializationError, e.ClusterOperationError) as err:
+            # noqa: E501
             msg = f"Error creating cluster '{cluster_name}'"
             LOGGER.error(msg, exc_info=True)
             if rollback:
@@ -847,6 +848,9 @@ class ClusterService(abstract_broker.AbstractBroker):
                               message=msg,
                               error_message=str(err))
         finally:
+            # TODO re-organize updating defined entity and task update as per
+            # https://stackoverflow.com/questions/49099637/how-to-determine-if-an-exception-was-raised-once-youre-in-the-finally-block
+            # noqa: E501
             self.context.end()
 
     @thread_utils.run_async
@@ -948,6 +952,9 @@ class ClusterService(abstract_broker.AbstractBroker):
                               message=msg,
                               error_message=str(err))
         finally:
+            # TODO re-organize updating defined entity and task update as per
+            # https://stackoverflow.com/questions/49099637/how-to-determine-if-an-exception-was-raised-once-youre-in-the-finally-block
+            # noqa: E501
             self.context.end()
 
     @thread_utils.run_async
@@ -1147,6 +1154,9 @@ class ClusterService(abstract_broker.AbstractBroker):
                               message=msg,
                               error_message=str(err))
         finally:
+            # TODO re-organize updating defined entity and task update as per
+            # https://stackoverflow.com/questions/49099637/how-to-determine-if-an-exception-was-raised-once-youre-in-the-finally-block
+            # noqa: E501
             self.context.end()
 
     def _get_cluster_upgrade_plan(self, source_template_name,
@@ -1353,6 +1363,9 @@ class ClusterService(abstract_broker.AbstractBroker):
                               message=msg, error_message=str(err))
 
         finally:
+            # TODO re-organize updating defined entity and task update as per
+            # https://stackoverflow.com/questions/49099637/how-to-determine-if-an-exception-was-raised-once-youre-in-the-finally-block
+            # noqa: E501
             self.context.end()
 
     @thread_utils.run_async
@@ -1416,6 +1429,9 @@ class ClusterService(abstract_broker.AbstractBroker):
                               message=msg,
                               error_message=str(err))
         finally:
+            # TODO re-organize updating defined entity and task update as per
+            # https://stackoverflow.com/questions/49099637/how-to-determine-if-an-exception-was-raised-once-youre-in-the-finally-block
+            # noqa: E501
             self.context.end()
 
     @thread_utils.run_async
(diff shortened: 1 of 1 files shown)


sakthisunda (approved):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/877)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Anirudh9794)

sahithi (changes requested):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/877)* status: all files reviewed, 1 unresolved discussion (waiting on @Anirudh9794)

___
*[container_service_extension/def_/cluster_service.py, line 1403 at r1](https://reviewable.io/reviews/vmware/container-service-extension/877#-MRw2QFz6B04ueuKKelg:-MRw2QFz6B04ueuKKelh:b-ic4n6f) ([raw file](https://github.com/vmware/container-service-extension/blob/c826df8b5c20ca6b9bbf893e83aef3378194b219/container_service_extension/def_/cluster_service.py#L1403)):*
> ```Python
>     def _fail_operation_and_resolve_entity(self, cluster_id: str,
>                                            op: DefEntityOperation,
>                                            should_resolve_entity=True,
> ```

I think the default should be False.
There is only one workflow (create) that attempts to resolve the entity. Every other operation doesn't need resolve.

Anirudh9794:
___
*[container_service_extension/def_/cluster_service.py, line 1403 at r1](https://reviewable.io/reviews/vmware/container-service-extension/877#-MRw2QFz6B04ueuKKelg:-MRwMH7b1C1hfbOxO1NV:b-896fix) ([raw file](https://github.com/vmware/container-service-extension/blob/c826df8b5c20ca6b9bbf893e83aef3378194b219/container_service_extension/def_/cluster_service.py#L1403)):*

I think the default should be False.
There is only one workflow (create) that attempts to resolve the entity. Every other operation doesn't need resolve.
</blockquote></details>

Done.

sahithi (approved):
:lgtm: 

> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/877#-:-MRwtUO66LHcZsv8PK5u:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @sahithi)

rocknes (changes requested):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/877)* status: all files reviewed, 2 unresolved discussions (waiting on @Anirudh9794 and @sahithi)

___
*[container_service_extension/def_/cluster_service.py, line 1397 at r2](https://reviewable.io/reviews/vmware/container-service-extension/877#-MRxwl4pAyaftD9_SYqu:-MRxwl4qCVWW_Bdseawz:bg6gz7w) ([raw file](https://github.com/vmware/container-service-extension/blob/e91094fcafd55f050ac3e13f68caf08966aafd95/container_service_extension/def_/cluster_service.py#L1397)):*
> ```Python
>         return self.entity_svc.update_entity(cluster_id, curr_entity)
> 
>     def _fail_operation_and_resolve_entity(self, cluster_id: str,
> ```

whenever a method has and it's name it generally means it is doing more than 1 distinct operation and these operations can be separated out in two different methods.

In this case it seems that resolving the entity is reserved for only one workflow, why not call the resolve endpoint only from that workflow and get rid of the bool and resolving logic from this method?

Anirudh9794:
___
*[container_service_extension/def_/cluster_service.py, line 1397 at r2](https://reviewable.io/reviews/vmware/container-service-extension/877#-MRxwl4pAyaftD9_SYqu:-MS4nGxAC2dBeCI0zsc6:bb6un1j) ([raw file](https://github.com/vmware/container-service-extension/blob/e91094fcafd55f050ac3e13f68caf08966aafd95/container_service_extension/def_/cluster_service.py#L1397)):*

whenever a method has and it's name it generally means it is doing more than 1 distinct operation and these operations can be separated out in two different methods.

In this case it seems that resolving the entity is reserved for only one workflow, why not call the resolve endpoint only from that workflow and get rid of the bool and resolving logic from this method?
</blockquote></details>

Done

rocknes (changes requested):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/877)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @Anirudh9794, @rocknes, and @sahithi)

___
*[container_service_extension/def_/cluster_service.py, line 822 at r3](https://reviewable.io/reviews/vmware/container-service-extension/877#-MS6FZujDn6yek5brroR:-MS6FZuk5SM-xTmxl2Ij:b-588g1v) ([raw file](https://github.com/vmware/container-service-extension/blob/36ab2d2ded8b6dd099beae3e2f6fa116a8dc8957/container_service_extension/def_/cluster_service.py#L822)):*
> ```Python
>             msg = f"Unknown error creating cluster '{cluster_name}'"
>             LOGGER.error(msg, exc_info=True)
>             # TODO: Avoid many try-except block. Check if it is a good practice
> ```

The newly added block of code is exactly the same for the specific except block and the general except block.

Maybe we can set a boolean in those except blocks that an exception was raised and then at the end of the method check the bool and perform the steps that you have added here. Does that sound better?

Btw I have no issues if we keep the code as is.

rocknes (approved):
:lgtm:

> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/877#-:-MS6R46b3jmduAR4E3yt:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @Anirudh9794 and @sahithi)

Anirudh9794:
___
*[container_service_extension/def_/cluster_service.py, line 822 at r3](https://reviewable.io/reviews/vmware/container-service-extension/877#-MS6FZujDn6yek5brroR:-MS9XDZl9FnQ8V0VQJIP:b-xeizxr) ([raw file](https://github.com/vmware/container-service-extension/blob/36ab2d2ded8b6dd099beae3e2f6fa116a8dc8957/container_service_extension/def_/cluster_service.py#L822)):*

The newly added block of code is exactly the same for the specific except block and the general except block.

Maybe we can set a boolean in those except blocks that an exception was raised and then at the end of the method check the bool and perform the steps that you have added here. Does that sound better?

Btw I have no issues if we keep the code as is.
</blockquote></details>

Added TODO for code reorganization as suggested. This should be done for both updating the task and setting the final state of the defined entity.

PR #951, opened 2021-03-08
[VCDA-2189] Update cluster delete to handle exposed cluster

ltimothy7:
Signed-off-by: ltimothy <ltimothy@vmware.com>

-Upon cluster delete, if the cluster is exposed, the dnat rule is deleted.
-Tested dnat rule successfully being deleted upon cluster delete for an exposed cluster.

@arunmk @rocknes @Anirudh9794

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/951)

### DIFF
--- a/container_service_extension/def_/cluster_service.py
+++ b/container_service_extension/def_/cluster_service.py
@@ -750,6 +750,8 @@ class ClusterService(abstract_broker.AbstractBroker):
                         network_name=network_name,
                         cluster_name=cluster_name,
                         cluster_id=cluster_id)
+                    LOGGER.info(f'Deleted dnat rule for cluster {cluster_name}'
+                                f'({cluster_id})')
 
                 try:
                     # Delete the corresponding defined entity
--- a/container_service_extension/def_/cluster_service.py
+++ b/container_service_extension/def_/cluster_service.py
@@ -344,6 +344,7 @@ class ClusterService(abstract_broker.AbstractBroker):
         cluster_name: str = curr_entity.name
         org_name: str = curr_entity.entity.metadata.org_name
         ovdc_name: str = curr_entity.entity.metadata.ovdc_name
+        network_name: str = curr_entity.entity.spec.settings.network
         phase: DefEntityPhase = DefEntityPhase.from_phase(
             curr_entity.entity.status.phase)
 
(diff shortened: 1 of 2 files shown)


arunmk (approved):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/951)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @ltimothy7)

___
*[container_service_extension/def_/cluster_service.py, line 746 at r1](https://reviewable.io/reviews/vmware/container-service-extension/951#-MVJCJ7XA9SBht7EoQ1H:-MVJCJ7XA9SBht7EoQ1I:b-9tzb9s) ([raw file](https://github.com/vmware/container-service-extension/blob/96343e485566ae9a8ad6164147517048052f99a9/container_service_extension/def_/cluster_service.py#L746)):*
> ```Python
> 
>                 if expose_ip:
>                     _handle_delete_expose_dnat_rule(
> ```

Maybe this needs to be in a try-except like the vapp above, so that we can continue with delete even if one fails.
___
*[container_service_extension/def_/cluster_service.py, line 1111 at r1](https://reviewable.io/reviews/vmware/container-service-extension/951#-MVJCQaI24QGPaHAl9lN:-MVJCQaI24QGPaHAl9lO:bi28k3a) ([raw file](https://github.com/vmware/container-service-extension/blob/96343e485566ae9a8ad6164147517048052f99a9/container_service_extension/def_/cluster_service.py#L1111)):*
> ```Python
> 
>             # Handle deleting dnat rule is cluster is exposed
>             if def_entity and def_entity.entity.status.exposed:
> ```

We may want to add try-catch around each delete so that we can continue deletion is any intermediate deletion fails. However this is more of a policy decision. In the current case it seems to be the right thing to do IMO.

ltimothy7 (commented):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/951)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @arunmk and @rocknes)

___
*[container_service_extension/def_/cluster_service.py, line 746 at r1](https://reviewable.io/reviews/vmware/container-service-extension/951#-MVJCJ7XA9SBht7EoQ1H:-MVJHmDXCS5MJd1k8Ftc:b-hr08w0) ([raw file](https://github.com/vmware/container-service-extension/blob/96343e485566ae9a8ad6164147517048052f99a9/container_service_extension/def_/cluster_service.py#L746)):*
<details><summary><i>Previously, arunmk (Arun M. Krishnakumar) wrote…</i></summary><blockquote>

Maybe this needs to be in a try-except like the vapp above, so that we can continue with delete even if one fails.
</blockquote></details>

Done. Thank you, Arun.
___
*[container_service_extension/def_/cluster_service.py, line 1111 at r1](https://reviewable.io/reviews/vmware/container-service-extension/951#-MVJCQaI24QGPaHAl9lN:-MVJHqZ1Bar16pqkun3k:b-u2ozjr) ([raw file](https://github.com/vmware/container-service-extension/blob/96343e485566ae9a8ad6164147517048052f99a9/container_service_extension/def_/cluster_service.py#L1111)):*
<details><summary><i>Previously, arunmk (Arun M. Krishnakumar) wrote…</i></summary><blockquote>

We may want to add try-catch around each delete so that we can continue deletion is any intermediate deletion fails. However this is more of a policy decision. In the current case it seems to be the right thing to do IMO.
</blockquote></details>

Done. I agree; @rocknes should we change the task status if the dnat rule does not delete successfully?

PR #965, opened 2021-03-23
[VCDA-2274] Fix upgrade issue related to missing rights in `cse service account`

rocknes:
CSE service account is missing the 'Edit' right for RightsBundles, as a result, it is unable to publish rights bundle to tenant orgs.
This workflow will break any upgrade from CSE 2.6 or prior to CSE 3.0/CSE 3.1.

Additionally there is a logic bug around how CSE publishes the right bundle to tenants. In the current implementation, CSE doesn't preserve the list of tenants who already has the right bundle published to them. This PR fixes the bug.

Additional small fixes to 
* improve flake8 compliance
* change the flag for cse create-service-role to follow standard key to ignore ssl certificate verification

Testing:
ToDo

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/965)

### DIFF
--- a/container_service_extension/common/constants/server_constants.py
+++ b/container_service_extension/common/constants/server_constants.py
@@ -528,6 +528,7 @@ CSE_SERVICE_ROLE_RIGHTS = [
     "Right: Manage",
     "Right: View",
     "Rights Bundle: View",
+    "Rights Bundle: Edit",
     "Role: Create, Edit, Delete, or Copy",
     "Service Configuration: Manage",
     "Service Configuration: View",
--- a/container_service_extension/installer/right_bundle_manager.py
+++ b/container_service_extension/installer/right_bundle_manager.py
@@ -58,8 +58,12 @@ class RightBundleManager():
             f"{CloudApiResource.RIGHT_BUNDLES}/{right_bundle_id}" \
             "/tenants/publish"
         payload = {
-            "values": [self.cloudapi_client.get_org_urn_from_id(org_id)
-                       for org_id in org_ids]}
+            "values": [
+                {"id": self.cloudapi_client.get_org_urn_from_id(org_id)}
+                for org_id in org_ids
+            ]
+        }
+
         return self.cloudapi_client.do_request(
             method=shared_constants.RequestMethod.POST,
             cloudapi_version=CloudApiVersion.VERSION_1_0_0,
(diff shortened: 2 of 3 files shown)


Anirudh9794:
___
*[container_service_extension/installer/right_bundle_manager.py, line 62 at r1](https://reviewable.io/reviews/vmware/container-service-extension/965#-MWRhUSVCkbViCZZOJJ3:-MWRhUSVCkbViCZZOJJ4:b9xkj21) ([raw file](https://github.com/vmware/container-service-extension/blob/01dce6e56fddca620b1b1667ec93a07bae95c0fa/container_service_extension/installer/right_bundle_manager.py#L62)):*
> ```Python
>         shared_to_orgs = set()
>         page_num = 0
>         while True:
> ```

I just checked the 10.2 api explorer and tried out the POST equivalent. It adds the rights only to the specified orgs. Do you think it is a good idea to use it instead of iterating over multiple REST calls ? (Asking because we were already hit with a similar OVDC bug)

rocknes (commented):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/965)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @Anirudh9794)

___
*[container_service_extension/installer/right_bundle_manager.py, line 62 at r1](https://reviewable.io/reviews/vmware/container-service-extension/965#-MWRhUSVCkbViCZZOJJ3:-MWViY4-CohSNiVAUbBz:b-ify5h2) ([raw file](https://github.com/vmware/container-service-extension/blob/01dce6e56fddca620b1b1667ec93a07bae95c0fa/container_service_extension/installer/right_bundle_manager.py#L62)):*

I just checked the 10.2 api explorer and tried out the POST equivalent. It adds the rights only to the specified orgs. Do you think it is a good idea to use it instead of iterating over multiple REST calls ? (Asking because we were already hit with a similar OVDC bug)
</blockquote></details>

Thanks for pointing this out. Done :)

sahithi (approved):
:lgtm: 

> Reviewed 2 of 3 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/965#-:-MWVrG4867tL4S5Qnw-m:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @Anirudh9794)

Anirudh9794 (approved):
:lgtm:

> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/965#-:-MWX9eHw-asSKyb_tLtE:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Anirudh9794)

PR #966, opened 2021-03-23
[VCDA-2184] Client-connect-to-vcd-using-computed-api-version

sakthisunda:
- Installer and upgrader to use vcd client to use computed api version 
- Remove usage of config['vcd']['version']
- In this fix vcd clients will not use api_version in vcd section. Instead, use dynamically computed version of the api.
- Pass around api version to avoid repeating max version
- Updated the supported versions in config validator that has final fitered versions based on legacy_mode flag
- Manually tested upgrader, cse check, cse template and cse install to use the computed api version

@Anirudh9794 @rocknes @sahithi

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/966)

### DIFF
--- a/container_service_extension/installer/config_validator.py
+++ b/container_service_extension/installer/config_validator.py
@@ -234,6 +234,8 @@ def get_validated_config(config_file_name,
             [float(x) for x in config['service']['supported_api_versions']
                 if float(x) >= 35.0]
     config['vcd']['api_version'] = str(max(supported_api_versions_float))
+    # Update the supported versions
+    config['service']['supported_api_versions'] = supported_api_versions_float
 
     # Store telemetry instance id, url and collector id in config
     # This steps needs to be done after api_version has been computed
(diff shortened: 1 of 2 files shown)


rocknes (changes requested):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/966)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @sakthisunda)

___
*[container_service_extension/installer/config_validator.py, line 238 at r1](https://reviewable.io/reviews/vmware/container-service-extension/966#-MWVDPdjDrEkmP95OUwi:-MWVDPdkERNvCEJ6hDh9:b-1obfzu) ([raw file](https://github.com/vmware/container-service-extension/blob/34ec00c93fa449ea84899969e17a20f5113d8d9b/container_service_extension/installer/config_validator.py#L238)):*
> ```Python
>     config['vcd']['api_version'] = str(max(supported_api_versions_float))
>     # Update the supported versions
>     config['service']['supported_api_versions'] = supported_api_versions_float
> ```

I had added the following line in my last PR
```
config['vcd']['api_version'] = str(max(supported_api_versions_float))
```

not sure why this is missing here.

also config['service']['supported_api_versions'] has been set to the proper value 20 lines above.

PR #1076, opened 2021-06-21
[VCDA-2601] Create subdirectory under LOCAL SCRIPTS DIR

Anirudh9794:
… key during cse upgrade

* Create subdirectory under LOCAL_SCRIPTS_DIR to store script cache
* Ignore compute policy key during cse upgrade as it prevents updating template metadata once again after upgrade is complete.

Testing:
* Cse upgrade to check if ignoring keys work fine
* Template install and see if proper directory is created during cluster installation.

@rocknes @sahithi @ltimothy7 @sakthisunda @arunmk

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1076)

### DIFF
--- a/container_service_extension/common/constants/server_constants.py
+++ b/container_service_extension/common/constants/server_constants.py
@@ -291,6 +291,7 @@ class LocalTemplateKey(str, Enum):
     NAME = 'name'
     OS = 'os'
     REVISION = 'revision'
+    COOKBOOK_VERSION = 'cookbook_version'
     UPGRADE_FROM = 'upgrade_from'
     MIN_CSE_VERSION = 'min_cse_version'
     MAX_CSE_VERSION = 'max_cse_version'
@@ -473,6 +474,7 @@ class TemplateBuildKey(str, Enum):
     IP_ALLOCATION_MODE = 'ip_allocation_mode'
     STORAGE_PROFILE = 'storage_profile'
     CSE_PLACEMENT_POLICY = 'cse_placement_policy'
+    REMOTE_COOKBOOK_VERSION = 'remote_template_cookbook_version'
 
 
 @unique
--- a/container_service_extension/installer/configure_cse.py
+++ b/container_service_extension/installer/configure_cse.py
@@ -1570,6 +1570,7 @@ def _update_metadata_for_existing_templates(client: Client, config: dict,
             # metadata keys barring the catalog_item_name.
             # Add catalog_item_name to the dict remote_template_descriptor
             remote_template_descriptor[server_constants.LocalTemplateKey.CATALOG_ITEM_NAME] = catalog_item_name  # noqa: E501
+            remote_template_descriptor[server_constants.LocalTemplateKey.COOKBOOK_VERSION] = str(rtm.cookbook_version)  # noqa: E501
 
             # New keys to be added (min_cse_version and max_cse_version)
             # will already be in remote_template_descriptor.
@@ -1667,6 +1668,7 @@ def _process_existing_templates(client: Client, config: dict,
             client,
             catalog_name=catalog_name,  # noqa: E501
             org_name=org_name,
+            ignore_metadata_keys=[server_constants.LegacyLocalTemplateKey.COMPUTE_POLICY],  # noqa: E501
             legacy_mode=True,
             logger_debug=INSTALL_LOGGER)
     # update metadata with min_cse_version and max_cse_version for all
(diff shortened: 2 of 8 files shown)


sakthisunda (approved):
> Reviewed 7 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1076)* status: 7 of 8 files reviewed, all discussions resolved (waiting on @Anirudh9794)

rocknes (changes requested):
> Reviewed 8 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1076)* status: all files reviewed, 1 unresolved discussion (waiting on @Anirudh9794)

___
*[container_service_extension/installer/templates/local_template_manager.py, line 115 at r1](https://reviewable.io/reviews/vmware/container-service-extension/1076#-Mcro78y6pxRhUL0LucG:-Mcro78z1aUirddqBMrU:b-a72qar) ([raw file](https://github.com/vmware/container-service-extension/blob/130158744de69a8c6ba54087a6811e9cd0cec625/container_service_extension/installer/templates/local_template_manager.py#L115)):*
> ```Python
>         missing_metadata_keys = missing_metadata_keys - set(ignore_metadata_keys)  # noqa: E501
>         num_missing_metadata_keys = len(missing_metadata_keys)
>         if num_missing_metadata_keys == len(expected_metadata_keys):
> ```

After removing/ignoring some keys, will this condition be ever met?

Because expected key list will have the cookbook_version key.

Btw why do we want to hide this particular key? We are stamping it on the template, let it run through the whole system. If the need be the client can prune it.

Anirudh9794:
___
*[container_service_extension/installer/templates/local_template_manager.py, line 115 at r1](https://reviewable.io/reviews/vmware/container-service-extension/1076#-Mcro78y6pxRhUL0LucG:-Mctophb1wJAxkM3Kn-4:b-9ev0so) ([raw file](https://github.com/vmware/container-service-extension/blob/130158744de69a8c6ba54087a6811e9cd0cec625/container_service_extension/installer/templates/local_template_manager.py#L115)):*

After removing/ignoring some keys, will this condition be ever met?

Because expected key list will have the cookbook_version key.

Btw why do we want to hide this particular key? We are stamping it on the template, let it run through the whole system. If the need be the client can prune it.
</blockquote></details>

Before this change, all the templates (legacy and non-legacy) could have been read by calling `get_all_k8s_local_template_definition(legacy_mode=True)`.
After adding `cookbook_version` into template metadata, only legacy templates will be read with `get_all_k8s_local_template_definition(legacy_mode=True)`.

This will affect CSE upgrade because since we want CSE upgrade should be idempotent and hence we would want to read all the templates to update their metadata.

So i added the new parameter `ignore_metadata` to ignore `cookbook_version` only during cse upgrade

rocknes (approved):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1076)* status: all files reviewed, 1 unresolved discussion (waiting on @Anirudh9794)

___
*[container_service_extension/installer/templates/local_template_manager.py, line 115 at r1](https://reviewable.io/reviews/vmware/container-service-extension/1076#-Mcro78y6pxRhUL0LucG:-McxF6jqCZ5oGLFjRNT1:b-6dlple) ([raw file](https://github.com/vmware/container-service-extension/blob/130158744de69a8c6ba54087a6811e9cd0cec625/container_service_extension/installer/templates/local_template_manager.py#L115)):*

Before this change, all the templates (legacy and non-legacy) could have been read by calling `get_all_k8s_local_template_definition(legacy_mode=True)`.
After adding `cookbook_version` into template metadata, only legacy templates will be read with `get_all_k8s_local_template_definition(legacy_mode=True)`.

This will affect CSE upgrade because since we want CSE upgrade should be idempotent and hence we would want to read all the templates to update their metadata.

So i added the new parameter `ignore_metadata` to ignore `cookbook_version` only during cse upgrade
</blockquote></details>

This still sounds very hacky to me .. let me think over it a bit and get back to you. As a general rule of thumb, we should strive to avoid as much as possible, special casing in code.

rocknes (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1076#-:-MdNmn5kD5up5DN5zvNF:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Anirudh9794)

___
*[container_service_extension/installer/templates/local_template_manager.py, line 115 at r1](https://reviewable.io/reviews/vmware/container-service-extension/1076#-Mcro78y6pxRhUL0LucG:-MdNmToJByzBtVHd3f0d:b-30db1c) ([raw file](https://github.com/vmware/container-service-extension/blob/130158744de69a8c6ba54087a6811e9cd0cec625/container_service_extension/installer/templates/local_template_manager.py#L115)):*

This still sounds very hacky to me .. let me think over it a bit and get back to you. As a general rule of thumb, we should strive to avoid as much as possible, special casing in code.
</blockquote></details>

Gave it some thought and it seems we need the ignore_metadata param, but I am still not convinced that this check is valid.

PR #1171, opened 2021-08-23
[VCDA-2736] Get cluster config from rde

sakthisunda:
- Get Cluster config from kube config from RDE
- if not found in RDE get the config from control plane vm extra config
- Tested for Personae: Sys-admin, Cluster admin, Cluster author
- verified no more running scripts on vsphere
 
image
---
@arunmk @ltimothy7

### DIFF
--- a/container_service_extension/common/constants/shared_constants.py
+++ b/container_service_extension/common/constants/shared_constants.py
@@ -240,3 +240,10 @@ class ClusterDetailsKey(str, Enum):
 class HttpResponseHeader(str, Enum):
     LOCATION = 'Location'
     X_VMWARE_VCLOUD_TASK_LOCATION = 'X-VMWARE-VCLOUD-TASK-LOCATION'
+
+
+@unique
+class RDEProperty(str, Enum):
+    PRIVATE = 'private'
+    KUBE_TOKEN = 'kube_token'
+    KUBE_CONFIG = 'kube_config'
--- a/container_service_extension/rde/backend/cluster_service_2_x.py
+++ b/container_service_extension/rde/backend/cluster_service_2_x.py
@@ -22,7 +22,6 @@ import semantic_version as semver
 from container_service_extension.common.constants.server_constants import CLUSTER_ENTITY  # noqa: E501
 from container_service_extension.common.constants.server_constants import ClusterMetadataKey  # noqa: E501
 from container_service_extension.common.constants.server_constants import ClusterScriptFile, TemplateScriptFile  # noqa: E501
-from container_service_extension.common.constants.server_constants import CSE_CLUSTER_KUBECONFIG_PATH  # noqa: E501
 from container_service_extension.common.constants.server_constants import DefEntityOperation  # noqa: E501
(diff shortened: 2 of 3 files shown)


ltimothy7 (approved):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1171)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @sakthisunda)

arunmk (changes requested):
There is one more change needed: we should remove the lines in the script that copies the kubeconfig from /etc/kubernetes/admin.conf to /root/.kube

> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1171#-:-MhvMUaq6rUnZjDQGSki:beerhep)* status: 2 of 3 files reviewed, 3 unresolved discussions (waiting on @ltimothy7 and @sakthisunda)

___
*[container_service_extension/rde/backend/cluster_service_2_x.py, line 206 at r2](https://reviewable.io/reviews/vmware/container-service-extension/1171#-MhvK4y79QOkhNiECqhL:-MhvK4y79QOkhNiECqhM:bh5eyzn) ([raw file](https://github.com/vmware/container-service-extension/blob/fa3512b28214ebb8f390c2f66d21c6ed369bf530/container_service_extension/rde/backend/cluster_service_2_x.py#L206)):*
> ```Python
>                         shared_constants.RDEProperty.KUBE_CONFIG.value):
>             kube_config = curr_native_entity.status.private.kube_config
>         else:
> ```

Is this because of a race condition? We should not need to fallback to control plane VM since there may not be only one etc.
___
*[container_service_extension/rde/backend/cluster_service_2_x.py, line 2040 at r2](https://reviewable.io/reviews/vmware/container-service-extension/1171#-MhvLyYR-mCkHAX7kyY6:-MhvLyYR-mCkHAX7kyY7:bihvtii) ([raw file](https://github.com/vmware/container-service-extension/blob/fa3512b28214ebb8f390c2f66d21c6ed369bf530/container_service_extension/rde/backend/cluster_service_2_x.py#L2040)):*
> ```Python
>         # Form kubeconfig with internal ip
>         kubeconfig_with_exposed_ip = self._get_kube_config_from_rde(cluster_id)
>         if not kubeconfig_with_exposed_ip:
> ```

We should fail if this is not there rather than getting from VM.
___
*[container_service_extension/rde/backend/cluster_service_2_x.py, line 2075 at r2](https://reviewable.io/reviews/vmware/container-service-extension/1171#-Mhuz2TU49klELDVmJRV:-Mhuz2TU49klELDVmJRW:b-vobste) ([raw file](https://github.com/vmware/container-service-extension/blob/fa3512b28214ebb8f390c2f66d21c6ed369bf530/container_service_extension/rde/backend/cluster_service_2_x.py#L2075)):*
> ```Python
>         rde = self.entity_svc.get_entity(cluster_id)
>         native_entity: rde_2_x.NativeEntity = rde.entity
>         if hasattr(native_entity.status,
> ```

Can we have an explicit `return None` for the case where the if condition does not hold?

PR #1185, opened 2021-09-09
Miscellaneous work for TKGm R1 CPI/CSI integration

arunmk:
VCDA-2776: CSI and CPI integration with TKGm
    VCDA-2475: set disk.enableUUID reusing Sakthi's code
    VCDA-2796: pass vip subnet to scripts
    VCDA-2781: reload vapp before checking status
    VCDA-2869: don't use placement policy for TKGm
    Also has:
    - fail strongly when expose_ip fails
    - handle vm.reboot failures
    - make scripts log better
    - fix resolve to use the correct way for 20.04 and beyond

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1185)


arunmk:
@rocknes @sakthisunda @ltimothy7 I will merge this to unblock testing and raise a ticket to handle review comments.

PR #1279, opened 2022-01-07
[VCDA-3162] Update gateway fetching to allow for datacenter group networks

andrew-ni:
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1279)

### DIFF
--- a/container_service_extension/rde/backend/cluster_service_1_x.py
+++ b/container_service_extension/rde/backend/cluster_service_1_x.py
@@ -791,7 +791,6 @@ class ClusterService(abstract_broker.AbstractBroker):
                     expose_ip = nw_exp_helper.expose_cluster(
                         client=self.context.client,
                         org_name=org_name,
-                        ovdc_name=ovdc_name,
                         network_name=network_name,
                         cluster_name=cluster_name,
                         cluster_id=cluster_id,
@@ -921,7 +920,6 @@ class ClusterService(abstract_broker.AbstractBroker):
                         nw_exp_helper.handle_delete_expose_dnat_rule(
                             client=self.context.client,
                             org_name=org_name,
-                            ovdc_name=ovdc_name,
                             network_name=network_name,
                             cluster_name=cluster_name,
                             cluster_id=cluster_id)
@@ -1087,7 +1085,6 @@ class ClusterService(abstract_broker.AbstractBroker):
                     nw_exp_helper.handle_delete_expose_dnat_rule(
                         client=self.context.client,
                         org_name=org_name,
-                        ovdc_name=ovdc_name,
                         network_name=network_name,
                         cluster_name=cluster_name,
                         cluster_id=cluster_id)
@@ -1361,7 +1358,6 @@ class ClusterService(abstract_broker.AbstractBroker):
                     nw_exp_helper.handle_delete_expose_dnat_rule(
                         client=self.context.client,
                         org_name=org_name,
-                        ovdc_name=ovdc_name,
                         network_name=network_name,
                         cluster_name=cluster_name,
                         cluster_id=cluster_id)
--- a/container_service_extension/rde/backend/cluster_service_2_x.py
+++ b/container_service_extension/rde/backend/cluster_service_2_x.py
@@ -929,7 +929,6 @@ class ClusterService(abstract_broker.AbstractBroker):
                     expose_ip = nw_exp_helper.expose_cluster(
                         client=self.context.client,
                         org_name=org_name,
-                        ovdc_name=ovdc_name,
                         network_name=network_name,
                         cluster_name=cluster_name,
                         cluster_id=cluster_id,
@@ -1075,7 +1074,6 @@ class ClusterService(abstract_broker.AbstractBroker):
                         nw_exp_helper.handle_delete_expose_dnat_rule(
                             client=self.context.client,
                             org_name=org_name,
-                            ovdc_name=ovdc_name,
                             network_name=network_name,
                             cluster_name=cluster_name,
                             cluster_id=cluster_id)
@@ -1238,7 +1236,6 @@ class ClusterService(abstract_broker.AbstractBroker):
                     nw_exp_helper.handle_delete_expose_dnat_rule(
                         client=self.context.client,
                         org_name=org_name,
-                        ovdc_name=ovdc_name,
                         network_name=network_name,
                         cluster_name=cluster_name,
                         cluster_id=cluster_id)
@@ -1543,7 +1540,6 @@ class ClusterService(abstract_broker.AbstractBroker):
                         nw_exp_helper.handle_delete_expose_dnat_rule(
                             client=self.context.client,
                             org_name=org_name,
-                            ovdc_name=ovdc_name,
                             network_name=network_name,
                             cluster_name=cluster_name,
                             cluster_id=cluster_id)
--- a/container_service_extension/rde/backend/cluster_service_2_x_tkgm.py
+++ b/container_service_extension/rde/backend/cluster_service_2_x_tkgm.py
@@ -983,7 +983,6 @@ class ClusterService(abstract_broker.AbstractBroker):
                         nw_exp_helper.handle_delete_expose_dnat_rule(
                             client=self.context.client,
                             org_name=org_name,
-                            ovdc_name=ovdc_name,
                             network_name=network_name,
                             cluster_name=cluster_name,
                             cluster_id=cluster_id)
@@ -1150,7 +1149,6 @@ class ClusterService(abstract_broker.AbstractBroker):
                     nw_exp_helper.handle_delete_expose_dnat_rule(
                         client=self.context.client,
                         org_name=org_name,
-                        ovdc_name=ovdc_name,
                         network_name=network_name,
                         cluster_name=cluster_name,
                         cluster_id=cluster_id
@@ -1442,7 +1440,6 @@ class ClusterService(abstract_broker.AbstractBroker):
                         nw_exp_helper.handle_delete_expose_dnat_rule(
                             client=self.context.client,
                             org_name=org_name,
-                            ovdc_name=ovdc_name,
                             network_name=network_name,
                             cluster_name=cluster_name,
                             cluster_id=cluster_id)
@@ -2076,7 +2073,6 @@ def _add_control_plane_nodes(
                     expose_ip = nw_exp_helper.expose_cluster(
                         client=user_client,
                         org_name=org.get_name(),
-                        ovdc_name=vdc.name,
                         network_name=network_name,
                         cluster_name=cluster_name,
                         cluster_id=cluster_id,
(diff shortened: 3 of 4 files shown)


andrew-ni (commented):
A note regarding the API JSON response handling in _get_gateway() function:

I'm directly accessing the response body json as well as the response headers, and i'm using strings to do so. 

I think we should eventually have a standardized cloudapi json processing layer/process, where raw json responses would be serialized into the respective VCD object (Network here), and then all other code would just use the Network object. This way if VCD API makes a breaking change, the error & fix would be localized to that json processing layer, instead of in random functions like _get_gateway(). 

However, there seems to be a majority dependence on pyvcloud constructs which use query api (xml resources), so i'm not sure how necessary a processing layer + model classes would be. Either way I think that's out of scope, and this implementation is sufficient and avoids further code clutter. 

> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1279#-:-MsnR98rAnmKzAdFMr3-:b6l6dlj)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @arunmk and @rocknes)

ltimothy7 (approved):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1279)* status: all files reviewed, 1 unresolved discussion (waiting on @andrew-ni, @arunmk, and @rocknes)

___
*[container_service_extension/rde/backend/common/network_expose_helper.py, line 53 at r1](https://reviewable.io/reviews/vmware/container-service-extension/1279#-Msqo5tiA_JKCm0O-JjK:-Msqo5tiA_JKCm0O-JjL:b-q7bgvm) ([raw file](https://github.com/vmware/container-service-extension/blob/09e2c7c680ea06e0f070b87f5b468cd90e4ff12f/container_service_extension/rde/backend/common/network_expose_helper.py#L53)):*
> ```Python
>             is_target_network = entry['name'] == network_name and \
>                 entry['orgRef']['name'] == org_name and \
>                 entry['networkType'] == 'NAT_ROUTED'
> ```

For clarity, could you make a note that expose only works for routed networks to make it clear why we are checking the networkType?

rocknes (changes requested):
> Reviewed 3 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1279)* status: all files reviewed, 5 unresolved discussions (waiting on @andrew-ni and @arunmk)

___
*[container_service_extension/rde/backend/common/network_expose_helper.py, line 50 at r1](https://reviewable.io/reviews/vmware/container-service-extension/1279#-Mt4O5V67FfTRqK_K5UY:-Mt4O5V67FfTRqK_K5UZ:b-m6tiee) ([raw file](https://github.com/vmware/container-service-extension/blob/09e2c7c680ea06e0f070b87f5b468cd90e4ff12f/container_service_extension/rde/backend/common/network_expose_helper.py#L50)):*
> ```Python
>             resource_url_relative_path=base_path + f'{page}',
>             return_response_headers=True)
>         for entry in response['values']:
> ```

I don't think we need the for loop,

if count of results == 0 raise missing network exception
if count of results > 1 raise multiple record exception
if count == 1 process
___
*[container_service_extension/rde/backend/common/network_expose_helper.py, line 51 at r1](https://reviewable.io/reviews/vmware/container-service-extension/1279#-Mt4NC1K77xULte5pJbt:-Mt4NC1L8kJxHw4JLjv0:b-as07vh) ([raw file](https://github.com/vmware/container-service-extension/blob/09e2c7c680ea06e0f070b87f5b468cd90e4ff12f/container_service_extension/rde/backend/common/network_expose_helper.py#L51)):*
> ```Python
>             return_response_headers=True)
>         for entry in response['values']:
>             is_target_network = entry['name'] == network_name and \
> ```

Is this check required at all? We already have the name filter in the query.
___
*[container_service_extension/rde/backend/common/network_expose_helper.py, line 53 at r1](https://reviewable.io/reviews/vmware/container-service-extension/1279#-Mt4NcOB8bZE2Cr249NQ:-Mt4NcOB8bZE2Cr249NR:b-4uyn2o) ([raw file](https://github.com/vmware/container-service-extension/blob/09e2c7c680ea06e0f070b87f5b468cd90e4ff12f/container_service_extension/rde/backend/common/network_expose_helper.py#L53)):*
> ```Python
>             is_target_network = entry['name'] == network_name and \
>                 entry['orgRef']['name'] == org_name and \
>                 entry['networkType'] == 'NAT_ROUTED'
> ```

I think these equality checks can be moved to the query directly.
___
*[container_service_extension/rde/backend/common/network_expose_helper.py, line 56 at r1](https://reviewable.io/reviews/vmware/container-service-extension/1279#-Mt4NgkXAbvzhVpdDbFx:-Mt4NgkXAbvzhVpdDbFy:bvbl7la) ([raw file](https://github.com/vmware/container-service-extension/blob/09e2c7c680ea06e0f070b87f5b468cd90e4ff12f/container_service_extension/rde/backend/common/network_expose_helper.py#L56)):*
> ```Python
>             if is_target_network:
>                 if gateway_name is not None:
>                     raise MultipleRecordsException(f"Multiple routed networks named {network_name} found. CSE Server expects unique network names.")  # noqa: E501
> ```

If the result size i.e. response['count'] (or something similar) is > 1 then we should raise the Multiple record exception.

andrew-ni (commented):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1279)* status: all files reviewed, 5 unresolved discussions (waiting on @arunmk, @ltimothy7, and @rocknes)

___
*[container_service_extension/rde/backend/common/network_expose_helper.py, line 50 at r1](https://reviewable.io/reviews/vmware/container-service-extension/1279#-Mt4O5V67FfTRqK_K5UY:-Mt4PT3D-Jz9a00_-DMK:b-v56esf) ([raw file](https://github.com/vmware/container-service-extension/blob/09e2c7c680ea06e0f070b87f5b468cd90e4ff12f/container_service_extension/rde/backend/common/network_expose_helper.py#L50)):*

I don't think we need the for loop,

if count of results == 0 raise missing network exception
if count of results > 1 raise multiple record exception
if count == 1 process
</blockquote></details>

Due to problem of filter restrictions, I can't just rely on the response count, as there's some cases where multiple networks of the same name will be returned in the API response, but only 1 of those will be a valid network.

For example, I can have a datacenter group routed network named `network1` as well as an ovdc isolated network named `network1`. In this case, only the datacenter group routed network will be valid, but 2 results will be returned.

Another case would be if the same network name exists in multiple orgs and the user is sys admin, or another cross org user. In this case, only the network in the specified org will be valid, but multiple results can be returned
___
*[container_service_extension/rde/backend/common/network_expose_helper.py, line 51 at r1](https://reviewable.io/reviews/vmware/container-service-extension/1279#-Mt4NC1K77xULte5pJbt:-Mt4PQmoC3K6-eS-J6AW:b-fff5rp) ([raw file](https://github.com/vmware/container-service-extension/blob/09e2c7c680ea06e0f070b87f5b468cd90e4ff12f/container_service_extension/rde/backend/common/network_expose_helper.py#L51)):*

Is this check required at all? We already have the name filter in the query.
</blockquote></details>

you're right, i'll remove this name check
___
*[container_service_extension/rde/backend/common/network_expose_helper.py, line 53 at r1](https://reviewable.io/reviews/vmware/container-service-extension/1279#-Mt4NcOB8bZE2Cr249NQ:-Mt4Owor8ETyD3me3VsH:b-aevu6d) ([raw file](https://github.com/vmware/container-service-extension/blob/09e2c7c680ea06e0f070b87f5b468cd90e4ff12f/container_service_extension/rde/backend/common/network_expose_helper.py#L53)):*

I think these equality checks can be moved to the query directly.
</blockquote></details>

These equality checks cannot be moved directly to the query. VCD API forbids filtering on these parameters (bad request error). Because of that, we must do these checks 
___
*[container_service_extension/rde/backend/common/network_expose_helper.py, line 53 at r1](https://reviewable.io/reviews/vmware/container-service-extension/1279#-Msqo5tiA_JKCm0O-JjK:-Mt4PO5pEL_wbKqf1F6P:b-xgnzkd) ([raw file](https://github.com/vmware/container-service-extension/blob/09e2c7c680ea06e0f070b87f5b468cd90e4ff12f/container_service_extension/rde/backend/common/network_expose_helper.py#L53)):*

For clarity, could you make a note that expose only works for routed networks to make it clear why we are checking the networkType?
</blockquote></details>

Sure will do
___
*[container_service_extension/rde/backend/common/network_expose_helper.py, line 56 at r1](https://reviewable.io/reviews/vmware/container-service-extension/1279#-Mt4NgkXAbvzhVpdDbFx:-Mt4PAoUCJyKBSrXh5Qk:b4m2swc) ([raw file](https://github.com/vmware/container-service-extension/blob/09e2c7c680ea06e0f070b87f5b468cd90e4ff12f/container_service_extension/rde/backend/common/network_expose_helper.py#L56)):*

If the result size i.e. response['count'] (or something similar) is > 1 then we should raise the Multiple record exception.
</blockquote></details>

Due to above problem of filter restrictions, I can't just rely on the response count, as there's some cases where multiple networks of the same name will be returned in the API response. See other comment for use cases

rocknes (approved):
:lgtm: once the outstanding comment is resolved

> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1279#-:-Mt5DR6EDCbmMu3Fadp_:b-ti10jc)* status: all files reviewed, 2 unresolved discussions (waiting on @andrew-ni, @arunmk, and @ltimothy7)

___
*[container_service_extension/rde/backend/common/network_expose_helper.py, line 55 at r2](https://reviewable.io/reviews/vmware/container-service-extension/1279#-Mt5D2fWEyy9pZMb_kx4:-Mt5D2fX60LxFDnmBKSM:b-sc8wcp) ([raw file](https://github.com/vmware/container-service-extension/blob/3d302c2b9369dc86b96d573328bac166a4a08e1f/container_service_extension/rde/backend/common/network_expose_helper.py#L55)):*
> ```Python
>                 entry['networkType'] == 'NAT_ROUTED'
>             if is_target_network:
>                 if gateway_name is not None:
> ```

can you please use an explicit int/boolean e.g. network_match_count or found_target_network .. makes the code a bit clearer.

andrew-ni (commented):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1279)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @arunmk, @ltimothy7, and @rocknes)

___
*[container_service_extension/rde/backend/common/network_expose_helper.py, line 55 at r2](https://reviewable.io/reviews/vmware/container-service-extension/1279#-Mt5D2fWEyy9pZMb_kx4:-Mt5ELawDGif0wP4aUDj:b-896fix) ([raw file](https://github.com/vmware/container-service-extension/blob/3d302c2b9369dc86b96d573328bac166a4a08e1f/container_service_extension/rde/backend/common/network_expose_helper.py#L55)):*

can you please use an explicit int/boolean e.g. network_match_count or found_target_network .. makes the code a bit clearer.
</blockquote></details>

Done.

PR #1281, opened 2022-01-10
[VCDA-3192] Remove dependency on config file from the commands `cse template import`

rocknes:
* Made the -c/--config param optional
* Added -x/--extra-option new param, this param takes two strings as input (key-value), this param can be repeated
* In case --config option is not provided by user, the -x params are used to create a dummy config dictionary in memory and passed onto the following methods
* Every method in the template import workflow that depends on config has been redesigned to explicitly state the actual fields of config it uses. This helps immensely to narrow down the scope of the dummy config file.
* A minimum set of required fields have been defined, if all of them is not specified by -x option, CSE will prompt the user to enter them interactively.
* Additional care has been taken to make sure that type-casting is done properly to the input key-value (str, str) pair received via -x option.

Testing Done:
Ran the command `cse template import`
1. without -c and -x options. Got prompted for all the required details
2. without -c and partial -x options. Got prompted for only the missing details
3. without -c and all required -x options. No prompt, command executed as normal.

Signed-off-by: rocknes <aritra.iitkgp@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1281)

### DIFF
--- a/container_service_extension/common/utils/core_utils.py
+++ b/container_service_extension/common/utils/core_utils.py
@@ -196,7 +196,7 @@ def str_to_bool(s):
 
     :param s: input string
 
-    :return: True if val is ['true' or 'yes' or 'Y'] otherwise False
+    :return: True if val is ['true' or 'yes' or 'y'] otherwise False
     """
     return str(s).lower() in ('true', 'yes', 'y')
 
--- a/container_service_extension/common/utils/core_utils.py
+++ b/container_service_extension/common/utils/core_utils.py
@@ -90,9 +90,9 @@ def get_installed_cse_version() -> semantic_version.Version:
     return semantic_version.Version('.'.join(tokens))
 
 
-def prompt_text(text, color='black', hide_input=False):
+def prompt_text(text, color='black', hide_input=False, type=str):
     click_text = click.style(str(text), fg=color)
-    return click.prompt(click_text, hide_input=hide_input, type=str)
+    return click.prompt(click_text, hide_input=hide_input, type=type)
 
 
 def is_environment_variable_enabled(env_var_name):
@@ -196,9 +196,9 @@ def str_to_bool(s):
 
     :param s: input string
 
-    :return: True if val is 'true' otherwise False
+    :return: True if val is ['true' or 'yes' or 'Y'] otherwise False
     """
-    return str(s).lower() == 'true'
+    return str(s).lower() in ('true', 'yes', 'y')
 
 
 def get_sha256(filepath):
--- a/container_service_extension/installer/config_validator.py
+++ b/container_service_extension/installer/config_validator.py
@@ -50,7 +50,7 @@ from container_service_extension.lib.nsxt.nsxt_client import NSXTClient
 from container_service_extension.lib.pksclient.api_client import ApiClient
 from container_service_extension.lib.pksclient.configuration import Configuration  # noqa: E501
 from container_service_extension.lib.telemetry.telemetry_utils import \
-    store_telemetry_settings
+    update_with_telemetry_settings
 from container_service_extension.lib.uaaclient.uaaclient import UaaClient
 from container_service_extension.logging.logger import NULL_LOGGER
 from container_service_extension.logging.logger import SERVER_NSXT_WIRE_LOGGER
@@ -213,8 +213,14 @@ def get_validated_config(
         logger_debug=logger_debug
     )
 
-    config = _add_additional_details_to_config(
+    config = add_additional_details_to_config(
         config=config,
+        vcd_host=config['vcd']['host'],
+        vcd_username=config['vcd']['username'],
+        vcd_password=config['vcd']['password'],
+        verify_ssl=config['vcd']['verify'],
+        is_legacy_mode=config['service']['legacy_mode'],
+        is_mqtt_exchange=server_utils.should_use_mqtt_protocol(config),
         log_wire=log_wire,
         log_wire_file=log_wire_file
     )
@@ -260,36 +266,50 @@ def get_validated_config(
     return config
 
 
-def _add_additional_details_to_config(
-        config: Dict,
-        log_wire: bool,
-        log_wire_file: str
+def add_additional_details_to_config(
+    config: Dict,
+    vcd_host: str,
+    vcd_username: str,
+    vcd_password: str,
+    verify_ssl: bool,
+    is_legacy_mode: bool,
+    is_mqtt_exchange: bool,
+    log_wire: bool,
+    log_wire_file: str
 ):
     """Update config dict with computed key-value pairs.
 
     :param dict config:
+    :param str vcd_host:
+    :param str vcd_username:
+    :param str vcd_password:
+    :param bool verify_ssl:
+    :param bool is_legacy_mode:
+    :param bool is_mqtt_exchange:
     :param bool log_wire:
     :param str log_wire_file:
 
     :return: the updated config file
     :rtype: dict
     """
-    is_legacy_mode = config['service']['legacy_mode']
-
     # Compute common supported api versions by the CSE server and vCD
     sysadmin_client = None
     try:
         sysadmin_client = Client(
-            config['vcd']['host'],
-            verify_ssl_certs=config['vcd']['verify'],
+            vcd_host,
+            verify_ssl_certs=verify_ssl,
             log_file=log_wire_file,
             log_requests=log_wire,
             log_headers=log_wire,
-            log_bodies=log_wire)
-        sysadmin_client.set_credentials(BasicLoginCredentials(
-            config['vcd']['username'],
-            SYSTEM_ORG_NAME,
-            config['vcd']['password']))
+            log_bodies=log_wire
+        )
+        sysadmin_client.set_credentials(
+            BasicLoginCredentials(
+                vcd_username,
+                SYSTEM_ORG_NAME,
+                vcd_password
+            )
+        )
 
         vcd_supported_api_versions = \
             set(sysadmin_client.get_supported_versions_list())
@@ -326,10 +346,19 @@ def _add_additional_details_to_config(
     config['service']['default_api_version'] = max_vcd_api_version_supported
     config['service']['rde_version_in_use'] = semantic_version.Version(
         rde_utils.get_runtime_rde_version_by_vcd_api_version(
-            max_vcd_api_version_supported))
+            max_vcd_api_version_supported
+        )
+    )
 
     # Update the config dict with telemetry specific key value pairs
-    store_telemetry_settings(config)
+    update_with_telemetry_settings(
+        config_dict=config,
+        vcd_host=vcd_host,
+        vcd_username=vcd_username,
+        vcd_password=vcd_password,
+        verify_ssl=verify_ssl,
+        is_mqtt_exchange=is_mqtt_exchange
+    )
 
     return config
 
--- a/container_service_extension/installer/configure_cse.py
+++ b/container_service_extension/installer/configure_cse.py
@@ -44,7 +44,7 @@ from container_service_extension.lib.telemetry.telemetry_handler \
 from container_service_extension.lib.telemetry.telemetry_handler import \
     record_user_action_details
 from container_service_extension.lib.telemetry.telemetry_utils import \
-    store_telemetry_settings
+    update_with_telemetry_settings
 from container_service_extension.logging.logger import INSTALL_LOGGER
 from container_service_extension.logging.logger import INSTALL_WIRELOG_FILEPATH
 from container_service_extension.logging.logger import NULL_LOGGER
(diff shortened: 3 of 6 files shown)


ltimothy7 (changes requested):
> Reviewed 2 of 5 files at r1, 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1281)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @Anirudh9794, @arunmk, @rocknes, and @sahithi)

___
*[container_service_extension/server/cli/server_cli.py, line 1605 at r3](https://reviewable.io/reviews/vmware/container-service-extension/1281#-Mt9QDTp477co1gB0TZY:-Mt9QDTp477co1gB0TZZ:bf6v9a0) ([raw file](https://github.com/vmware/container-service-extension/blob/95b4b4cb2a0d1c08964b12cfdfd8c4c21bcddcbe/container_service_extension/server/cli/server_cli.py#L1605)):*
> ```Python
> 
>     # Prompt user for missing required options
>     for k, option in required_options.items():
> ```

I think it would be more straightforward if this function didn't also do the prompting. I think the caller function should check if all of the required options are added. Then, this function can just construct the config.

This would avoid adding the user input again after the additional values are retrieved.

rocknes (commented):
> Reviewed 4 of 5 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1281)* status: all files reviewed, 1 unresolved discussion (waiting on @Anirudh9794, @arunmk, @ltimothy7, and @sahithi)

___
*[container_service_extension/server/cli/server_cli.py, line 1605 at r3](https://reviewable.io/reviews/vmware/container-service-extension/1281#-Mt9QDTp477co1gB0TZY:-Mt9RSByBuoXwSLqCnw-:bkry04c) ([raw file](https://github.com/vmware/container-service-extension/blob/95b4b4cb2a0d1c08964b12cfdfd8c4c21bcddcbe/container_service_extension/server/cli/server_cli.py#L1605)):*

I think it would be more straightforward if this function didn't also do the prompting. I think the caller function should check if all of the required options are added. Then, this function can just construct the config.

This would avoid adding the user input again after the additional values are retrieved.

</blockquote></details>

Makes sense, I will refactor. Thanks.

sahithi (commented):
> Reviewed 3 of 5 files at r1, 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1281)* status: all files reviewed, 5 unresolved discussions (waiting on @Anirudh9794, @arunmk, @ltimothy7, and @rocknes)

___
*[container_service_extension/server/cli/server_cli.py, line 1295 at r3](https://reviewable.io/reviews/vmware/container-service-extension/1281#-Mt9m2HpB4dEYwOUdMid:-Mt9m2HqEJQJ01EtWND0:bsfkr59) ([raw file](https://github.com/vmware/container-service-extension/blob/95b4b4cb2a0d1c08964b12cfdfd8c4c21bcddcbe/container_service_extension/server/cli/server_cli.py#L1295)):*
> ```Python
>     type=(str, str),
>     multiple=True,
>     help='Extra options passed as -x key value'
> ```

Help string can be more descriptive on when/why this needs to be used? and with the combination of what other options?
___
*[container_service_extension/server/cli/server_cli.py, line 1303 at r3](https://reviewable.io/reviews/vmware/container-service-extension/1281#-Mt9l-Yo0OyAKdC1e6Q8:-Mt9l-Yo0OyAKdC1e6Q9:bv4anhs) ([raw file](https://github.com/vmware/container-service-extension/blob/95b4b4cb2a0d1c08964b12cfdfd8c4c21bcddcbe/container_service_extension/server/cli/server_cli.py#L1303)):*
> ```Python
>         ova_file_path,
>         force_import,
>         extra_options
> ```

If "extra options" can prompt users with "required options" messages, then I suggest renaming.
How about "upload_options" ?
___
*[container_service_extension/server/cli/server_cli.py, line 1307 at r3](https://reviewable.io/reviews/vmware/container-service-extension/1281#-Mt9mkhx7Y1Lv1UDWAfG:-Mt9mkhx7Y1Lv1UDWAfH:b8h4inx) ([raw file](https://github.com/vmware/container-service-extension/blob/95b4b4cb2a0d1c08964b12cfdfd8c4c21bcddcbe/container_service_extension/server/cli/server_cli.py#L1307)):*
> ```Python
>     """Upload TKGm OVA to CSE catalog.
> 
>     \b
> ```

Can you also add a few examples of the command usage?
___
*[container_service_extension/server/cli/server_cli.py, line 1312 at r3](https://reviewable.io/reviews/vmware/container-service-extension/1281#-Mt9g__c5zaQd5V3l_xn:-Mt9g__c5zaQd5V3l_xo:bhpximj) ([raw file](https://github.com/vmware/container-service-extension/blob/95b4b4cb2a0d1c08964b12cfdfd8c4c21bcddcbe/container_service_extension/server/cli/server_cli.py#L1312)):*
> ```Python
>             imported.
>         broker.org: String: Name of the Organization where the OVA will be
>             imported.
> ```

I understand that you named the keys in such a way that we can construct the config dict out of it easily.
It is more user-intuitive to use the term "catalog" rather than "broker.catalog". Users may begin to wonder what the broker is all about.

Can we use simpler keys and let the code deal with the needed retrofitting?

Another option is to take a yaml input.  It becomes easier for users to populate the yaml all at once and reuse it. I know it's a bit late for this suggestion, but I think it is worth considering. We can provide the sample yaml either in the command help string (or) thru sample option

vcd:
..host:
..password:
..username:
..org:
..catalog:
ova_file:



_Code quote:_
```Python
        broker.catalog: String: Name of the catalog where the OVA will be
            imported.
        broker.org: String: Name of the Organization where the OVA will be
            imported.
        vcd.host: String: FQDN of VCD host.
        vcd.password: String: Password of CSE service account.
        vcd.username: String: Username of CSE service account.
```

rocknes (commented):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1281)* status: all files reviewed, 5 unresolved discussions (waiting on @Anirudh9794, @arunmk, @ltimothy7, @rocknes, and @sahithi)

___
*[container_service_extension/server/cli/server_cli.py, line 1295 at r3](https://reviewable.io/reviews/vmware/container-service-extension/1281#-Mt9m2HpB4dEYwOUdMid:-Mt9ovsz2sTTZu0Ltk5Q:bebrt1w) ([raw file](https://github.com/vmware/container-service-extension/blob/95b4b4cb2a0d1c08964b12cfdfd8c4c21bcddcbe/container_service_extension/server/cli/server_cli.py#L1295)):*

Help string can be more descriptive on when/why this needs to be used? and with the combination of what other options?
</blockquote></details>

Will do
___
*[container_service_extension/server/cli/server_cli.py, line 1303 at r3](https://reviewable.io/reviews/vmware/container-service-extension/1281#-Mt9l-Yo0OyAKdC1e6Q8:-Mt9p0leEZ_AvPdXNFLa:b9iziuv) ([raw file](https://github.com/vmware/container-service-extension/blob/95b4b4cb2a0d1c08964b12cfdfd8c4c21bcddcbe/container_service_extension/server/cli/server_cli.py#L1303)):*

If "extra options" can prompt users with "required options" messages, then I suggest renaming.
How about "upload_options" ?
</blockquote></details>

`extra options` is an industry standard though, generally used in bash.
___
*[container_service_extension/server/cli/server_cli.py, line 1307 at r3](https://reviewable.io/reviews/vmware/container-service-extension/1281#-Mt9mkhx7Y1Lv1UDWAfG:-Mt9p7gv7egUZAa2WPlN:bhy9ot6) ([raw file](https://github.com/vmware/container-service-extension/blob/95b4b4cb2a0d1c08964b12cfdfd8c4c21bcddcbe/container_service_extension/server/cli/server_cli.py#L1307)):*

Can you also add a few examples of the command usage?
</blockquote></details>

Will do.
___
*[container_service_extension/server/cli/server_cli.py, line 1312 at r3](https://reviewable.io/reviews/vmware/container-service-extension/1281#-Mt9g__c5zaQd5V3l_xn:-Mt9pDZ_98S-8S4OQWXP:b-tz5ytc) ([raw file](https://github.com/vmware/container-service-extension/blob/95b4b4cb2a0d1c08964b12cfdfd8c4c21bcddcbe/container_service_extension/server/cli/server_cli.py#L1312)):*

I understand that you named the keys in such a way that we can construct the config dict out of it easily.
It is more user-intuitive to use the term "catalog" rather than "broker.catalog". Users may begin to wonder what the broker is all about.

Can we use simpler keys and let the code deal with the needed retrofitting?

Another option is to take a yaml input.  It becomes easier for users to populate the yaml all at once and reuse it. I know it's a bit late for this suggestion, but I think it is worth considering. We can provide the sample yaml either in the command help string (or) thru sample option

vcd:
..host:
..password:
..username:
..org:
..catalog:
ova_file:



</blockquote></details>

I will add a second layer of display key to yaml key mapping, the change should be simple and straight forward. 
I really want to avoid accepting another yaml. It just makes the automation tedious and complicated.

rocknes (commented):
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1281)* status: all files reviewed, 5 unresolved discussions (waiting on @Anirudh9794, @arunmk, @ltimothy7, @rocknes, and @sahithi)

___
*[container_service_extension/server/cli/server_cli.py, line 1303 at r3](https://reviewable.io/reviews/vmware/container-service-extension/1281#-Mt9l-Yo0OyAKdC1e6Q8:-Mt9pmLXDe0a1ZUxMQLa:b4g6328) ([raw file](https://github.com/vmware/container-service-extension/blob/95b4b4cb2a0d1c08964b12cfdfd8c4c21bcddcbe/container_service_extension/server/cli/server_cli.py#L1303)):*

`extra options` is an industry standard though, generally used in bash.
</blockquote></details>

The use case is as follows,

If you don't specify the config file, you will get prompts but you want to automate the command, in that case, you specify the inputs via the --extra-option param.

The prompting is the primary behavior, extra options allows you to avoid the prompting.

sahithi (approved):
:lgtm: one question
What happens if the user misses out on providing all the required options using -x?

> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1281#-:-MtBha6H18NENlFLKorM:b-blx7an)* status: all files reviewed, 5 unresolved discussions (waiting on @Anirudh9794, @arunmk, @ltimothy7, @rocknes, and @sahithi)

___
*[container_service_extension/server/cli/server_cli.py, line 1303 at r3](https://reviewable.io/reviews/vmware/container-service-extension/1281#-Mt9l-Yo0OyAKdC1e6Q8:-MtBCecc-jj67yOGKLry:bpdrd5k) ([raw file](https://github.com/vmware/container-service-extension/blob/95b4b4cb2a0d1c08964b12cfdfd8c4c21bcddcbe/container_service_extension/server/cli/server_cli.py#L1303)):*

The use case is as follows,

If you don't specify the config file, you will get prompts but you want to automate the command, in that case, you specify the inputs via the --extra-option param.

The prompting is the primary behavior, extra options allows you to avoid the prompting.

</blockquote></details>

Got it. 
What happens if the user misses out on providing all the required options using -x?

rocknes (commented):
The command will prompt and ask for the missing options.

> Reviewed 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1281#-:-MtCGqJs4YoYMwDLs4Cq:b-x8utwp)* status: all files reviewed, 5 unresolved discussions (waiting on @Anirudh9794, @arunmk, @ltimothy7, @rocknes, and @sahithi)

___
*[container_service_extension/server/cli/server_cli.py, line 1303 at r3](https://reviewable.io/reviews/vmware/container-service-extension/1281#-Mt9l-Yo0OyAKdC1e6Q8:-MtCH0Ur4VJy2as9ZYIo:b-x8utwp) ([raw file](https://github.com/vmware/container-service-extension/blob/95b4b4cb2a0d1c08964b12cfdfd8c4c21bcddcbe/container_service_extension/server/cli/server_cli.py#L1303)):*

Got it. 
What happens if the user misses out on providing all the required options using -x?
</blockquote></details>

The command will prompt and ask for the missing options.

ltimothy7 (approved):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1281)* status: all files reviewed, 5 unresolved discussions (waiting on @Anirudh9794, @arunmk, @ltimothy7, and @sahithi)

PR #1307, opened 2022-02-11
[VCDA-3193] Add REST end point and CLI command to reload templates in server runtime config

rocknes:
Added a new endpoint POST /api/cse/templates/action/reload,
and a new CLI command `vcd cse template reload`

Both these operations are restricted to only sys-admins.

Testing Done:
Ran the command and the REST call as sys admin and monitored the resulting task which tracked the reload of templates.
Confirmed from CSE server debug log that the reload operation was successful.

---
This change is [Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1307)


sakthisunda (approved):
> Reviewed 5 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/vmware/container-service-extension/1307)* status: 5 of 9 files reviewed, all discussions resolved (waiting on @sakthisunda)

